PR #4, opened 2021-04-27
KOGITO-4868: SceSim scroll fix

treblereel:
See [KOGITO-4868](https://issues.redhat.com/browse/KOGITO-4868)

### DIFF
--- a/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/editor/ScenarioSimulationEditorKogitoWrapper.java
+++ b/drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/editor/ScenarioSimulationEditorKogitoWrapper.java
@@ -19,11 +19,18 @@ package org.drools.workbench.screens.scenariosimulation.kogito.client.editor;
 import java.util.List;
 import java.util.function.Supplier;
 
+import javax.annotation.PostConstruct;
 import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 
 import com.google.gwt.user.client.ui.IsWidget;
 import com.google.gwt.user.client.ui.PopupPanel;
+import elemental2.dom.CSSProperties;
+import elemental2.dom.DomGlobal;
+import elemental2.dom.Element;
+import elemental2.dom.HTMLCollection;
+import elemental2.dom.HTMLDivElement;
+import elemental2.dom.HTMLTableElement;
 import elemental2.promise.Promise;
 import jsinterop.base.Js;
 import org.drools.scenariosimulation.api.model.AbstractScesimData;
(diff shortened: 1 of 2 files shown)


caponetto (approved):
Thanks, @treblereel!

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/editor/ScenarioSimulationEditorKogitoWrapper.java
```diff
@@ -133,6 +139,22 @@ public ScenarioSimulationEditorKogitoWrapper(
...
+    @PostConstruct
+    public void init() {
+        scenarioSimulationEditorPresenter.getView().onResize();
+        HTMLCollection<Element> elements = DomGlobal.document.getElementsByClassName("tab-content");
+        if(elements.length == 1) {
+            HTMLDivElement element = (HTMLDivElement) elements.getAt(0);
+            element.style.height = CSSProperties.HeightUnionType.of("calc(100vh - 36px)");
+        }
+
+        DomGlobal.document.body.onresize = p0 -> {
```
yesamer, line 151: @treblereel Is it necessary to put that handler logic on `body` tag? I tried to put it on the `element` you defined and the fix seems working as well. I think we should avoid putting that handler in `body`, if possible.

yesamer, line 151: @treblereel Is it necessary to put that handler logic on `body` tag? I tried to put it on the `element` you defined and the fix seems working as well. I think we should avoid putting that handler in `body`, if possible.

yesamer (commented):
Thank you @treblereel, I have a question for you:

drools-wb-screens/drools-wb-scenario-simulation-editor/drools-wb-scenario-simulation-editor-kogito-client/src/main/java/org/drools/workbench/screens/scenariosimulation/kogito/client/editor/ScenarioSimulationEditorKogitoWrapper.java
yesamer: @treblereel It seems working to me even removing that `onResize` assignment, WDTY?

treblereel: @yesamer i need it for vscode and testing app, but yes, we can remove it in case we are focused on vscode

yesamer (approved):
@treblereel Thank you, LGTM

PR #16, opened 2021-04-30
Kogito-4980 : Stunner - Palette fixes & improvements

inodeman:
Hi @romartin this PR Fixes

1) Issue with Double clicking an icon while in the palette. Once detected the event will be canceled
2) Issue with node text appearing behind the palette in edit mode
3) Order of Custom Tasks

4) Issue about throwing Exception while dragging Text Anotation was fixed on my initial work with Data Objects
What is pending is the blurring of the icon while draggon but that can only be fixed with higher resolution images. Let me know what you want me to do.

Artifact can be found here. https://drive.google.com/drive/folders/1m7EQ-8zzc8MRzBQVV8kU2xUdnTg6bvFp?usp=sharing

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/test/java/org/kie/workbench/common/stunner/client/widgets/inlineeditor/InlineTextEditorBoxViewImplTest.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-widgets/src/test/java/org/kie/workbench/common/stunner/client/widgets/inlineeditor/InlineTextEditorBoxViewImplTest.java
@@ -84,6 +84,9 @@ public class InlineTextEditorBoxViewImplTest {
 
     private InlineTextEditorBoxViewImpl tested;
 
+    @Mock
+    private org.jboss.errai.common.client.dom.HTMLElement parentElement;
+
     @Before
     @SuppressWarnings("unchecked")
     public void init() {
@@ -91,6 +94,7 @@ public class InlineTextEditorBoxViewImplTest {
         this.tested.init(presenter);
         when(editNameBox.getStyle()).thenReturn(editNameBoxStyle);
         when(nameField.getStyle()).thenReturn(nameFieldStyle);
+        when(editNameBox.getParentElement()).thenReturn(parentElement);
         doAnswer(i -> {
             ((Scheduler.ScheduledCommand) i.getArguments()[0]).execute();
             return null;


domhanak:
jenkins run fdb

domhanak:
@inodeman same here, please rebase to that we get FBD :)

domhanak:
*Issue with Double clicking an icon while in the palette. Once detected the event will be canceled*
I can still reproduce this all thought it is not exactly a double-click.
1. Click on some anchor in palette, move a bit and click on the same anchor.
2. Item is in the palette.
What I expected is that item to be reset, same as when I do double-click on same anchor on palette. Could we improve that?

domhanak:
@LuboTerifaj https://issues.redhat.com/browse/KOGITO-4980 please check the linked issues for which you are tester.

romartin:
Hey @domhanak 

Good catch! We had been looking at your comment above and it makes sense, also testing what you mean it other tools, and it behaves different as well. So I think that's not just taking some decision right now about this behavior on top of palette, let's move it to a new ticket, so we can properly discuss it. WDYT? 

Thanks!

domhanak:
Hi @romartin the issue is not double click on palette, the issue is that when you click with a selection over an element that is over canvas the selection is pasted into canvas anyway and disregards where the user clicked. That's why I still think this is just a workaround for the root cause of the issue with palette and not a fix at all.
When you do a double click and move the mouse just one pixel in between the clicks the issue is back again.

Anyway reported here: https://issues.redhat.com/browse/KOGITO-5143

LuboTerifaj:
Hey @domhanak , @inodeman , the fixes are only for Kogito, right? However, the linked jiras are not. What exactly should I check? Shall I report the separate jiras also for Kogito? Thanks!

romartin:
oh good point @domhanak @LuboTerifaj 

This palette improvements are considered only a kogito specific improvement, so this why the main ticekt is a kogito one: https://issues.redhat.com/browse/KOGITO-4980

On the other hand, we're taking this opportunity to work on palette fixes & improvements in roder to fix some other small issues, reported time ago, so those are not really reported as KOGITO, they're old... anyway, I think it's good to include all those in our reviews and ensure we're fixing as much as possible on the palette. 

So @LuboTerifaj , what I can do is moving linked ticket to the KOGITO project, but then it should be @domhanak the tester I guess. In fact , I don't mind who is the reviewer guys, the key point is fixing as much as possible in this task, and so we included some older ticket as well. WDYT?

LuboTerifaj:
@domhanak , @inodeman , The reason why I am asking is whether the issues https://issues.redhat.com/browse/JBPM-9495 and https://issues.redhat.com/browse/RHPAM-3172 are supposed  to be fixed in this PR. Thanks!

domhanak:
Verified this PR, checked that following works:
1) When double clicking on the Palette, the event will be canceled/consumed.
DMN & BPMN :heavy_check_mark: 
2) The order of custom tasks is not sorted when building the palette
BPMN :heavy_check_mark: 
3) When editing text using the inline editor will show behind Properties panel or Expanded Palette
DMN & BPMN :heavy_check_mark: 


@LuboTerifaj the issues were just related for reference, and this RP fixes those for kogito editors. Apologies for not being clear enough. I figured the linked issue with @inodeman so all good now.

domhanak (approved):

romartin:
Hey @LuboTerifaj @domhanak 

Thanks for looking at the tickets. So finally the idea is that the main ticket:

https://issues.redhat.com/browse/KOGITO-4980

Its acceptance criteria already take into account the linked tickets. So actually this PR should fix the ticket and the linked ones as well.

Anyway thanks for the review and approval! If any Qs please reach us out.

Thanks!

romartin (approved):

PR #23, opened 2021-05-06
KOGITO-5082: Restore the notification mechanism

caponetto:
In this PR, the notification mechanism from [appformer](https://github.com/kiegroup/appformer/) is restored.

All necessary classes were copied as-is from appformer, 
though some minor changes had to be made in order to make the code compatible with this repository. 

Please keep in mind that this notification mechanism will probably be replaced in favor of each channel's approach.
For now, let's use this code so things are kept the way it was.

Example: Error notification is shown when loading an invalid dmn file in a scesim file
Screenshot from 2021-05-06 16-28-54

**Note**: This code only adds the infrastructure for showing the notification popup. No notification message has been changed.


tiagobento (approved):
Bringing this back while the Editors don't use Notifications API. Thanks @caponetto!

tomasdavidorg (approved):
Looks good to me.

PR #37, opened 2021-05-10
KOGITO-4872: Show an error page for SceSim in case `setContent` fails

caponetto:
See [KOGITO-4872](https://issues.redhat.com/browse/KOGITO-4872)

The `ErrorPage` widget, which has been used by stunner, has been moved to a common place so that scesim can reuse it as well.

**Result**:
KOGITO-4872


tiagobento (approved):

yesamer (approved):
Very nice, thank you @caponetto

domhanak (approved):
Thanks. Works well.

PR #38, opened 2021-05-12
Bump version to `8.1.0-SNAPSHOT`

karreiro:


yesamer (approved):

caponetto (approved):
Thanks, @karreiro!

I'm thinking about adding a script to make it easy for us to bump versions, like [this one](https://github.com/errai/errai/blob/master/updateVersions.sh).
Wdyt? @karreiro @tiagobento

karreiro:
@caponetto it seems like a great idea! Also, I would add an extra step for create the commit with a standard message (like "`Bump version to < newversion>`". Wdyt? :-)

PR #39, opened 2021-05-12
Bump version script and workflow

caponetto:
Script and workflow to make it easy to bump versions.
The workflow takes the `new_version` as input and creates a pull request with the changes.

PR #46, opened 2021-05-18
Workflow for creating the release branch

karreiro:
This PR introduces a workflow analog to the "Bump version workflow" (https://github.com/kiegroup/kogito-editors-java/pull/39). However, in this case it creates the release branch automatically.

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -3,6 +3,9 @@ name: "Bump Version"
 on:
   workflow_dispatch:
     inputs:
+      release_version:
+        description: "Release version"
+        required: true
       new_version:
         description: "New version"
         required: true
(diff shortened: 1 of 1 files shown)


caponetto (approved):
Automation FTW! Thanks, @karreiro!

tiagobento:
@karreiro Should this Workflow also update the version to the release version? I mean, when branching out from `main`, you'll get a SNAPSHOT version, which will require you to manually update the version after the branch is created, no? So to create a branch that's ready for the release, we could probably do everything in one Workflow. WDYT?

tiagobento:
@karreiro Thanks! I think you could use the @kogito-tooling-bot user to perform commits.

karreiro:
Thank you for the re-review, @caponetto @tiagobento Changes applied :-)

caponetto (approved):
Thanks, @karreiro! :)

tiagobento (approved):

PR #55, opened 2021-05-27
[DO NOT MERGE] KOGITO-5214: DMN Language Server API implemented with a mock to let editor developers work before it is actually completed

adrielparedes:
DO NOT MERGE.

This Mock is intended to be used for development purposes by @yesamer. We don't need to merge it until we have a more polished and final version.

This API contains a mock that will let DMN Editor developer start working with the API before the Language Server is complete.

### DIFF
--- a/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/ls/dmn/DMNLanguageServiceApi.java
+++ b/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/ls/dmn/DMNLanguageServiceApi.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2021. Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.appformer.kogito.bridge.client.ls.dmn;
+
+import java.util.List;
+
+import jsinterop.annotations.JsType;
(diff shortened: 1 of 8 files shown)


appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/ls/dmn/DMNLanguageServiceApiWrapper.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public native List<LanguageServiceResult> getClasses(String completion);
+
+    @Override
+    public native List<LanguageServiceResult> getMethods(String fqdn, String completion);
+
+    @Override
+    public native List<LanguageServiceResult> getAttributes(String fqdn, String completion);
+
+    @JsProperty(name = "dmnLanguageServiceApi")
```
tiagobento, line 37: Hi Adriel! NIce PR! Maybe this should be called `dmnLanguageServerApi` instead? Asking because from the PR title, the name of this new API is going to be `Language Server API`, right?

tiagobento: Also, I know this will only be used by DMN for now, but maybe we don't need to have the `dmn` prefix there?

yesamer: +1, considering the idea is to have a Language Server available for all the editors, I agree with @tiagobento to remove all `dmn` prefixes

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/ls/dmn/DMNLanguageServiceApi.java
```diff
@@ -0,0 +1,32 @@
...
+
+@JsType(isNative = true)
+public interface DMNLanguageServiceApi {
+
+    List<LanguageServiceResult> getClasses(String completion);
+
+    List<LanguageServiceResult> getMethods(String fqdn, String completion);
+
+    List<LanguageServiceResult> getAttributes(String fqdn, String completion);
+}
```
yesamer, line 32: @adrielparedes I suggest adding another method, which tells the API user if the language server is available or not.
Eg.
`boolean isLanguagueServerAvaliable()`.
This because we need to enable or disable a button based on that logic.

PR #58, opened 2021-06-08
KOGITO-4043: The nodes should be created on top of the selected node in DMN editor

danielzhe:
See: https://issues.redhat.com/browse/KOGITO-4043

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/canvas/controls/actions/DMNCreateNodeAction.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/canvas/controls/actions/DMNCreateNodeAction.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.workbench.common.dmn.client.canvas.controls.actions;
+
+import javax.enterprise.context.Dependent;
+import javax.enterprise.event.Event;
+import javax.enterprise.inject.Any;
+import javax.inject.Inject;
+
+import org.jboss.errai.ioc.client.api.ManagedInstance;
+import org.kie.workbench.common.dmn.api.definition.model.BusinessKnowledgeModel;
+import org.kie.workbench.common.dmn.api.definition.model.Decision;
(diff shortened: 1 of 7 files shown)


danielzhe:
jenkins do fdb

jomarko (commented):
@danielzhe hello, thank you for the PR.

I found only one issue. It seems during computing a slot for new Decision/BKM node, we firstly check a free slot according to the axis X. What causes results described below. From my point of view they are not expected. What do you think please?

State before adding node
Screenshot from 2021-06-10 06-46-28

After node is added to InputData
Screenshot from 2021-06-10 06-46-43


I would expect Decision-4 to be vertically aligned with Decision-2

karreiro (commented):
Thank you, @danielzhe. Great stuff!

@jomarko, thank you for the comment as well! While I understand the scenario you've highlighted seems a bit annoying. In this case, users are inverting the tree that we're trying to encourage, by using inputs living above of decisions. Still, I don't think we will be able to handle all possible scenarios. Let's see this example:

Screenshot 2021-06-11 at 11 58 16


In this case, we have many decision nodes at a lower level (considering "InputData-1" as the reference). In this kind of use case, only the business logic will help us to organize the diagram better.

Thus, I believe may stick with the simpler implementation that @danielzhe proposes, by creating the decisions always in the `RightUp` direction.

--

@danielzhe, I've notice an intermittent issue.

In the following gif we can see it buggy:

A

..but on this one we can see it working:

B


Wdyt?

Thanks again for this PR!

danielzhe:
Thank you for the review, @karreiro ! About this issue, I noticed that if you drag the mouse a little bit (1 pixel is enough) after clicking on the "Create DMN Decision", it happens exactly what you reported. So for me it seems it is not an issue related to this PR. Can you confirm, please?

About what you found, @jomarko, I agree that it can be a bit annoying but as Guilherme mentioned, we trying to encourage users to create trees putting decisions above inputs, using a simple logic. Thank you for the review!

karreiro:
@danielzhe I believe we may invest sometime to fix that issue, as I was able to frequently spot it locally :( 

Also, for some reason, I could not reproduce it on ~master~ main:

Kapture 2021-06-11 at 16 42 58

jomarko:
@danielzhe @karreiro ok, thank you, then I think we need to resolve just the issue @karreiro is reporting https://github.com/kiegroup/kogito-editors-java/pull/58#issuecomment-859633435

danielzhe:
Hi, @karreiro ! As we talked in private chat, I think this issue you reported is related to the speed of the click on the button and we are able to reproduce it on main even in BPMN. Let's wait now for @romartin review. Wdyt?

danielzhe:
Here is the JIRA for the issue found by @karreiro : https://issues.redhat.com/browse/KOGITO-5328

karreiro (approved):
Thank you, @danielzhe.

jomarko (approved):
Thank you @danielzhe

romartin (approved):
Looks good to me as well, thanks @danielzhe !

romartin:
@danielzhe @karreiro before merging, could someone double check BPMN works as before plz? I see some code updated on the core, although looks good at first glance, but would be nice having someone checking BPMN doesn't get broken. Also another option is assigning someone from QE to test it. Or if you push the vsix or war, I can also quickly test locallly.

karreiro:
Thanks, @romartin! I just double-checked BPMN locally, and it's fine 🚀  (here's the [plugin](https://www.dropbox.com/s/sh8extec9826w98/KOGITO-4043.vsix?dl=0) in case you wanna double as well) :-)

romartin:
@danielzhe @karreiro all good, go ahead guys , thanks!

PR #72, opened 2021-07-12
KOGITO-5137: Conscious language initiative

tiagobento:
Adjusting our codebase confirming to the conscious language initiative.

@romartin Can you please take a closer look at some constants named `ROOT_PATH`? They seem like something important :)

Some references to other kiegroup repositories might not work until they update their default branch to `main`.


romartin:
Good catch @tiagobento , I see we have a reference in the code to VFS (the `ROOT_PATH` stuff). It also same thing done in the DMN side.

So @karreiro , you know if we can just drop this code about hardcoded paths in our editors? As far as I see looking at code ,it's only used here [1] in BPMN, and that diagram Metadata I think makes no sense for the kogito stream (it still does on kie-wb stream, but I don't see any usage here)

[1] https://github.com/kiegroup/kogito-editors-java/blob/38cb11270ce519b960f24cc4a2fa98b7e21304ba/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-marshalling/src/main/java/org/kie/workbench/common/stunner/bpmn/client/marshall/service/BPMNClientDiagramService.java#L191

WDYT? Can we just drop those references to vfs paths?

karreiro:
@romartin @tiagobento good catch! +1 for removing vfs paths from Kogito stream. I don't see any reason for keeping them neither.

tiagobento:
@romartin @karreiro Removed in 70d662b. Please check the commit individually to see if I didn't do anything wrong 😝 .

romartin (approved):
oh @tiagobento , nice job about dropping the root path stuff!! even interfaces! thanks a lot! :1st_place_medal: 

I don't see any issues here, so for me everything is good to go! Only @tiagobento , please if you can sure process & dmn diagrams can be just exported / imported properly, after removing root path stuff I mean, that would be great as well.

karreiro (approved):
Great stuff, @tiagobento! Thank you.

PR #73, opened 2021-07-14
KOGITO-1625: [DMN Designer] Data Types changes do not mark asset as dirty

danielzhe:
See https://issues.redhat.com/browse/KOGITO-1625

[VSCode Plugin](https://drive.google.com/file/d/1N2RIFKsysfYGiqmJxllhurDnSSpQ0qVe/view?usp=sharing)

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeList.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/types/listview/DataTypeList.java
@@ -24,6 +24,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import javax.annotation.PostConstruct;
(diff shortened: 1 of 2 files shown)


karreiro (approved):
Great stuff 🚀 thank you, @danielzhe!

manstis:
FWIW, you might want to be careful here. Using the `CommandManager` does indeed mark the editor as dirty since the VSCode mechanism relies on the command stack to know if there have been changes made in the editor. However that also means these changes can now be _undone_. This on it's own is no bad thing... but IIRC the rest of the Data Type Editor does not support undo/redo and hence by supporting it now you're opening a door to be asked to add it elsewhere.

karreiro:
Hi @manstis! Yes, you're correct! :-) However, given the efforts that the team is performing the Data Dictionary, we're being a bit diligent about investing efforts in the current implementation of the Data Types editor.  So, we've followed for this more straightforward approach.

jomarko (commented):
@danielzhe thank you for the PR. I have no comments to code. During manual review I found this corner case that I am not sure about.

- Add a data type
- Confirm it by checkmark - asset marked as dirty - perfect
- Delete previously added data type using trash icon
- Asset is now in the same state as during opening time, however still marked as dirty, is that expected?

jomarko:
@manstis @danielzhe re https://github.com/kiegroup/kogito-editors-java/pull/73#issuecomment-881291258, I was not able to undo redo data types changes.

danielzhe:
Hi, @manstis  Since we are doing a new Data Type editor, we decided not to put much effort to change the current Data Type editor to support undo/redo and focused only in a change that prevent user from losing data for closing diagrams without being notified that there is pending changes.

About the issue you reported, @jomarko , we understand that the effort to track if the current deleted Data Type restore the diagram to the previous saved state (so the "is dirty" is cleared) doesn't compensate the benefit, so this corner case is expected.

jomarko (approved):
Thank you @danielzhe

PR #71, opened 2021-07-12
Kogito-5470 : Cannot import process

inodeman:
Hi @romartin applied changes 1) Enable Comments on XML 2) Enable empty variable name on Multiple Instance can you please review this PR?

PR #75, opened 2021-07-15
KOGITO-5506 : BPMN Editor - Marshallers encoding issues

inodeman:
Hi @romartin as discussed, removing calls to replaceIllegalChars, since gwt classes do that now. Can you review this PR?

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/DataObject.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-api/src/main/java/org/kie/workbench/common/stunner/bpmn/definition/DataObject.java
@@ -41,7 +41,6 @@ import org.kie.workbench.common.stunner.core.util.HashUtil;
 
 import static org.kie.workbench.common.forms.adf.engine.shared.formGeneration.processing.fields.fieldInitializers.nestedForms.AbstractEmbeddedFormsInitializer.COLLAPSIBLE_CONTAINER;
 import static org.kie.workbench.common.forms.adf.engine.shared.formGeneration.processing.fields.fieldInitializers.nestedForms.AbstractEmbeddedFormsInitializer.FIELD_CONTAINER_PARAM;
-import static org.kie.workbench.common.stunner.core.util.StringUtils.revertIllegalCharsAttribute;
 
 @Portable
 @Bindable
@@ -88,7 +87,7 @@ public class DataObject extends BaseArtifacts {
 
         super(backgroundSet, fontSet, dimensionsSet);
         this.name = name;
-        this.name.setValue(revertIllegalCharsAttribute(this.name.getValue()));
+        this.name.setValue(this.name.getValue());
         this.type = type;
         this.general = general;
     }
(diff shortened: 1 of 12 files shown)


romartin:
Hey @inodeman 

Thanks for the fix, it looks good to me.

Although it's actually hard to know if we're breaking something by this change, as actually, on this kogito stream, we have not many marshalling tests (as we have in the "old" kie-wb-common )

So please, I think in order to ensure we don't break "marshallers" stuff, until we don't hve new marshallers and unit tests, it would be good idea "moving" these fixes to the "old" kie-wb-common stream, and execute all unit tests on that repo. As far I see the codebse for kie-wb-common and kogito, at marshallers level, should be similar, so I think it worths moving our latest fixees to the marshallers being done in this repo (so also including https://github.com/kiegroup/kogito-editors-java/pull/71 ) to the kie-wb-common side. This way we'll be able to execute all test case and ensure we don't break anything being tested before. If the fixes finally also make sene to include in kie-wb-common, even not requested by the ticket itself, we can also consider PRing there and fixing that repo.

WDYT?

inodeman:
Hi @romartin, @domhanak I tested the issue on kie-wb-commons and the issue does not happen. Can you review this?

romartin:
Thanks for the updates @inodeman 

Please @hasys , added you as the review for engineering as you're also up to date on these issues. Please sync with @domhanak and @inodeman to get this ready to merge.

Thanks guys!

hasys (changes requested):
Hi @inodeman, I think issue not in the escaping some symbols, but that in some cases they are not escaping back.

I think we need to add `org.kie.workbench.common.stunner.core.util.StringUtils.revertIllegalCharsAttribute` method call there: https://github.com/kiegroup/kogito-editors-java/blob/80a34e97ae88541fcd884d76690f07fe11890479/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-marshalling/src/main/java/org/kie/workbench/common/stunner/bpmn/client/marshall/converters/tostunner/activities/BaseCallActivityConverter.java#L56

Did you check it?

inodeman:
> Hi @inodeman, I think issue not in the escaping some symbols, but that in some cases they are not escaping back.
> 
> I think we need to add `org.kie.workbench.common.stunner.core.util.StringUtils.revertIllegalCharsAttribute` method call there:
> 
> https://github.com/kiegroup/kogito-editors-java/blob/80a34e97ae88541fcd884d76690f07fe11890479/kie-wb-common-stunner/kie-wb-common-stunner-sets/kie-wb-common-stunner-bpmn/kie-wb-common-stunner-bpmn-marshalling/src/main/java/org/kie/workbench/common/stunner/bpmn/client/marshall/converters/tostunner/activities/BaseCallActivityConverter.java#L56
> 
> Did you check it?

Hi @hasys yes, I checked it. The issue is this. Our marshallers escape the text "Task & Char" to "Task &amp; Char" which is correct, the problem is that the XML saver from GWT then changes "Task &amp; Char" for "Task &amp;amp; Char" which is where the issue occurs, now when the diagram is loaded again, the characters are reverted so the text would be "Task &amp; Char" so it would appear that it did not revert illegal charecters but in fact it did.

hasys:
Hi @inodeman, great finding, thank you! Just one more question, every `StringUtils.replaceIllegalChars` should have respective `StringUtils.revertIllegalCharsAttribute` calls as well. Why didn't it help? And should we remove it now after `StringUtils.replaceIllegalChars`  is removed?

inodeman:
> Hi @inodeman, great finding, thank you! Just one more question, every `StringUtils.replaceIllegalChars` should have respective `StringUtils.revertIllegalCharsAttribute` calls as well. Why didn't it help? And should we remove it now after `StringUtils.replaceIllegalChars` is removed?

Hi @hasys you are right, I did some tests and seems the revert calls are not needed anymore.I will remove them

inodeman:
Hi @romartin @hasys applied requested changes, can you re-review?

hasys (approved):
Great job, thank you @inodeman!

domhanak:
Hello, checked the vscode extension and I can no longer reproduce the issue.

I used following process:
Screenshot from 2021-07-28 12-09-58


Special characters were not replaced when saved and reopened. :heavy_check_mark: 
Some examples for values in XML: (all other nodes are the same)
Value saved in XML `<drools:metaValue><![CDATA[Save & Test]]></drools:metaValue>` :heavy_check_mark: 
Value saved in XML `<drools:metaValue><![CDATA[end & error]]></drools:metaValue>` :heavy_check_mark: 
Vaue saved in XML `<drools:metaValue><![CDATA[Calculate %]]></drools:metaValue>` :heavy_check_mark: 

LGTM. Thanks

domhanak (approved):

domhanak:
@kiegroup/gatekeepers ready for merging

PR #76, opened 2021-07-20
KOGITO-5549 - WID files with comments and Imports can't be loaded

hasys:
Hi @romartin, @domhanak,

now it is possible to use comments (both single and multi line) and imports in WID files.

see [VS Code extension](https://drive.google.com/file/d/10tCqW6yruYsebGY9V-6JR88EMD6IGOcS/view?usp=sharing) for tests. Thank you!


romartin:
Hey @hasys 

Code looks good. Although the PR title still talks "imports", but I think we're just supporting comments here. Is that correct? Can we change the title in that case to specify what's strictly fixed?

Thx in advance!

romartin (approved):
Forget my comment above please. I see we're just skipping everything before object starts, good point. Everything ok.
Thanks!

domhanak:
Thanks I ll review tommorow

domhanak (approved):
LGTM

domhanak:
@kiegroup/gatekeepers ready to be merged.

PR #79, opened 2021-07-23
KOGITO-5571: No effects when assigning a not-expression Simple Type column to expression type (and viceversa)

yesamer:
**JIRA**: https://issues.redhat.com/browse/KOGITO-5571
@jomarko This PR fixes the 2^ issue you found here https://github.com/kiegroup/drools-wb/pull/1503#pullrequestreview-703931429 . Considering you found it, I assign the QE review to you, but please feel free to reassign.

Step to reproduce: 
- Create a new Test Scenario (RULE)
- Form "Data Object" tab, add any Java class available. (eg. Integer)
- Add the Integer > value property in the grid
- Try to re-assign the same column to an expression, selecting "Integer" class in the Test Tools.
- Nothing happens
- (the same occurs if you assign the expression first and then reassign it with Integer > value.

Same fix introduced in BC here: https://github.com/kiegroup/drools-wb/pull/1504


jomarko (commented):
@yesamer thank you for the PR, however I am not sure if it Is needed. Is there a Data Object tab in the VS Code extension?

yesamer:
@jomarko Yes, you're right, currently this seems not useful for Kogito, but the strategy is to keep Kogito and BC versions synchronized as more as possible. In this way, if in the future RULE test scenario will be enabled in Kogito, no fixes or functionality will be lost.

jomarko (approved):
@yesamer thank you, then approving as there is nothing to check currently from QE point of view.

danielzhe (approved):

PR #80, opened 2021-07-27
KOGITO-5091 - creating connection can't be cancelled easily

hasys:
Hi @romartin, @domhanak, @jstastny-cz,

there is the fix. Issue was escape for actions wasn't migrated to Kogito hot keys API correctly. I migrated it and removed old parts of the API which is not in use anymore.

Fix works for both BPMN and DMN editors. I tested not only my fix but also all shortkeys from "Keyboard shortcuts" on my laptop, all works good.

[Tested VSIX](https://drive.google.com/file/d/1XmAHnDNhiu8cCTr0SVdCDav4yTpWHBOP/view?usp=sharing)


jstastny-cz:
I'll leave code-reviews to others with deeper insight into the code-base. Just my personal opinion - the test removals seem to be fine with regards to refactoring made in this PR.

When testing the change using the attached vsix, I've encountered an issue with connector creation, see attached video. Is that a known issue or possibly a result of these changes?
connector-stuck

hasys:
Hi @jstastny-cz, thank you for quick review.

Good catch, however issue is reproducible on the master branch as well so I guess it shouldn't block the PR. Thank you!

jstastny-cz (approved):
Thanks for clarification, otherwise works fine, thank you.

romartin (approved):
Thanks @hasys , looks good to me! Good refactor as well :+1:

romartin:
Thanks @jstastny-cz , reported https://issues.redhat.com/browse/KOGITO-5600

PR #82, opened 2021-07-29
KOGITO-5481: JS - Solve performance gap (large boxed expressions takes too long to load - Relation 50x50) demo | branch - CSS / JS/TS/Hooks

vpellegrino:
## This PR is about two main changes:
 1. Improve the showcase application to enable it to receive as input (typed by the user) the Expression JSON definition and on-the-flight render the related Expression accordingly.
 2. A first incremental solution for improving rendering performance for large tables (e.g., `Relation` table) used in the Boxed Expression Editor
### Live demo
You can find it here: https://boxed-expression-editor.netlify.app/
### Technical details
In order to achieve the second goal, a series of improvements have been made:
 3. Optimizing the `EditableCell` component, making it a `PureComponent`, by using `React.memo` with a comparison function that has less effort as possible.
 4. Avoid `ExpressionContainer` from having a dedicated state prop for the expression: instead of receiving the expression definition as a prop directly from the `BoxedExpression` root component, and trigger a callback every time a change should be made.
 5. Optimizing every expression component by calling `window.beeApi` only when expression definition is effectively changed, for the subscribed properties. This means that rendering and reduced, and moreover, meaningless definition changes are not notified in the external world.
### Results
You can observe the results of optimization rendering directly in the GIF below:
*Before:*
before

*After:*
after

In the gif above, I followed the same steps and for sake of simplicity, there is a `Relation` table with one row and two columns, with a total of two editable cells. 
From React Component point of view, the hierarchy is: 
 6. `RelationExpression`
 7. `Table`
 8. `TableBody`
 9. `EditableCell`
Every time a component is rendered a `console.log` is triggered, reporting the name of the rendered component, plus some information about column and row, in the case of the `EditableCell`.
Flows are essentially two:
 1. Selecting the cell `column-1` - `row 0` and then selecting the cell `column-2` - `row 0`.
	 Before, while switching from a cell to another, the entire table, with all descendants (all cells included) were re-rendered.
	 After, re-rendering happens only for the cells involved in the selection/deselection operations, for making it visible to the user.
2. Starting from the cell `column-1` - `row 0`, edit it with a value ("a") and then switching to the cell  `column-2` - `row 0` with the tab key.
	Before, edit and switch operations were re-rendering `Table` with all descendants multiple times.
	After edit and switch operations is re-rendering `Table` only one time, avoiding the re-rendering of all cells, but only the cell that is going to be selected (`column-2` - `row 0`).
Overall, this means that every time all cells are re-rendered, in a context where a lot of cells are present (2500, in case of a table 50 x 50), the overload is sensitive.
However, as already stated, this is only a first optimization step, because as you can observe, in the case of huge tables, few perceptible delays are still observable when performing table operations, like resize a column, rename a column, add a row, etc. This is mainly because we need to assure data synchronization, that nested components work correctly, and that automatic operations happen as well (automatic resizing).
### Next steps
We still have several paths to explore for optimizing the editor:
- CSS optimizations (task not yet traced *CSS - Solve performance gap (large boxed expressions takes too long to load - Relation 50x50) demo | branch - CSS / JS/TS/Hooks*)
- Consider applying client-side pagination for avoiding the table to be too heavy of components at the same time.  `react-table` has some built-in support for paginating the rows. We should consider a typical use case to understand if it is feasible/makes sense to paginate the columns too.
- Further investigations for optimizing React components structure/data flow.


karreiro (approved):
Thank you, @vpellegrino.

jomarko (approved):
Thank you @vpellegrino

PR #85, opened 2021-07-30
KOGITO-5274: Stunner - Line splicing

handreyrc:
Hey @romartin, could you review this PR please?

**JIRA**: [KOGITO-5274](https://issues.redhat.com/browse/KOGITO-5274)
**VS Code**: [plugin](https://drive.google.com/file/d/1Adsc5IIyCZcM6RiM2Q4Yi5xwAxehuuHw/view?usp=sharing)


handreyrc:
jenkins retest this

handreyrc:
jenkins do fdb

handreyrc:
jenkins do cdb

romartin:
Hey @handreyrc ,

Thanks for the great work. 

As far as I know, there exist a coulpe of known issues. Although the overall work looks great, and it works good in most of the cases, could you please share with us (me and @domhanak ) both issues, so we don't have to report again, if facing with those?

Thanks in advance!

domhanak (changes requested):
Hello everyone:
If two nodes are connected with sequence flow and the third one is dragged and dropped over this sequence flow, a new one sequence flow is created:
* If the connection is allowed the connector is highlighted in blue when the node is being dragged over the connector. :warning: 
    * **Issue found - 01** - Does not work for Sub process activities. When you drag node over a sequence that is inside of (EmbeddedSubprocess, EventSubprocess, MultipleInstanceSubprocess) the sequence is colored **red** and the line splicing succeeds.
    The color of the sequence should be blue.
* Target of the initial sequence flow changes to a new ("dropped") node :heavy_check_mark: 
* The new sequence flow has the Source set to the new node and the Target of this flow is set to the Target of the initial sequence flow. :warning: 
    * **Issue found 02** - When you save and reopen the diagram on which you performed line splicing, the sequences do not use **central connection point** of the activity which results in ugly diagram drawing. See the picture. It does not matter if the sequence is inside of container or not. Issue manifest in both cases.
    
KOGITO-5274-saveReopenMalformed

     
* The new node is placed at the position it was dropped. :heavy_check_mark: 
* If the line splice is being performed in a container such as a lane, the containment operation has to be performed as well. :heavy_check_mark: 
* Undo and Redo operations are supported. :heavy_check_mark: 
* Line splice must be compliant to all allow node connection rules. :heavy_check_mark: 

Other cases:

* If the connection is not allowed the connector is highlighted in red when the node is being dragged over the connector. :warning: -> See issue 01
* In case there is no enough space to splice the initial sequence flow, the line splice isn't performed. :heavy_check_mark: 
    * Note: this is something that is not well testable as the `not enough space` is not defined anywhere. However using exploratory testing I did not find any evidence of bad behavior and the user experience was good for me.
* Multiple bend points on sequences handling :heavy_check_mark: 

Imho the issues 01 and 02 needs to be resolved before we can merge. Let me know and thanks for this awesome feature!

romartin:
Thanks a lot @domhanak for you review.

Yeah I think issue 01 is a "known issue", I mean, we already identified it. This bug IMO is not high priority at this point, cause rather than the connector gets in red color (which means that the action shoud not be allowed), it works, so at least the process authoring can continue. Anyway let's keep in in our radar for fixing.

About other issues, thanks for pointing those out. Could you please @handreyrc take a look?

Thanks guys!

handreyrc:
Hey @romartin, 
In regards with the issue 01, I agree with you. It's a highlight issue, something else is interfering with the highlight, however,  it does not break anything and the line splicing works.  Maybe we can open a separete Jira for this one.

* About the issue 02, great catch @domhanak! I was testing only in between tasks and it works fine. It is fixed already, I've just pushed the fix. 
 **VS Code**: [plugin](https://drive.google.com/file/d/1In1-pTkhyHbyZ057FHzrFSClyNI-aglt/view?usp=sharing)

There is one more known issue to keep in track. 
* Issue 03 : Containment issue when line splicing. Sometimes when the line splice is happening into a container (eg. lane/sub-process) the parentPicker control is not able to pick the parent properly and only the line splice is executed. If you move the container you will notice that the node was not added to the container.  It happens when the node precisely overlaps the connector in the exact point the parentPicker tries to get the parent. At some point we will h

I think we are good for a second round!

Thanks @romartin  @domhanak

domhanak:
@handreyrc Unfortunately, Issue 02 is not fully fixed:
Screenshot from 2021-08-05 12-51-45

It seems that not all pair combinations -> START(NODE_TYPE_A) -> END(NODE_TYPE_B) <- are handled.
I would like to see these combination tested in this PR as this is not possible to verify on E2E level.
Please reach out to me if this can not be tested and we can discuss other possiblities.

Please report issue 03 I am not able to reproduce so video or mode detailed description will help.

For issue 01 - it is highlight issue, but for users it is very tricky to know that the red color does not mean it is not allowed. They will be confused and raising question. Imho it should be fixed.

handreyrc:
Hey @domhanak,  

In regards with issues we have:

* Issue 02: (Connections) I alredy fixed the connection issues for the combinations you reported. It's ready to be retested.
**VS Code**: [plugin](https://drive.google.com/file/d/1FlgovTSpTz3U0nz_gVZr4I-G-IOrTxIK/view?usp=sharing)

* Issue 01: (Highlight) I found out why the highlight is red within the subprocess, actualy the highlight is correct based on the connection rules. There is rule for subprocesses where you cannot connect none of the internal nodes to any external node, it means that at the moment you are dragging the node from outside of the subprocess into it, that node desn't belong to the subprocess yet and the connector is highlited in red. However, when you drop the node the line splice is executed and the containment operation is executed too making it a valid operation and that's why it works. Considering a different scenario, if the node you want to splice is already inside the subprocess, you will see that the highlight works fine because the conainment was executed before the splice.

* Issue 03: There is an easy way to reproduce it. Drop a lane, add two connected tasks into the lane, drop a Gateway ouside the lane, then slowly drag the Gateway vertically over the middle of sequence flow inside the lane until the shadow around the lane that indicates the cotainment is allowed disapears, finally drop the shape. This is the point where the parentPicker control can't get the parent underneath the connector and only the splice will be executed. If you move the lane the Gateway will be moved below the lane. Let me know if you can't reproduce it.

Hopefully the Issue 02 is fixed and we can report the issues 01 and 03 separatedly and move forward with this PR.

Thank you so much @domhanak!
Cc @romartin

domhanak:
**Issue 01** - Makes sense, in that case it is behaving properly, so this is resolved. Thank you for the explanation. Maybe we can improve on this somehow in the future, but I do not have any suggestions now.

**Issue 02** - I was no longer able to reproduce this issues. Could you please link the unit tests testing this? I want to double check if I covered all cases. I got a bit lost in the code after force-push.

**Issue 03** - Yes I managed to reproduce, very nice catch. Could you please report it, as you are the one that found it?
To closely elaborate, the issue is that the NODE you are dragging is being dragged over a different node, in this case it is SEQUENCE FLOW in the lane so the EDITOR does not know you are above a Lane and forgets the containment operation. Moving the dragged node slightly away from above the sequence (designer thinks the node is above the LANE now) the containment happens.
You can check this, by creating a lane with `start->usertask-> end` nodes and hovering above these nodes with some other node. While hovering the LANE's shade (containment indicator) is not there. Once the hover moves away, the containment shade is back.

@handreyrc Please let me know about the tests for issue 02 and please ping me about issue 03. I can approve after these questions are resolved..

kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/test/java/org/kie/workbench/common/stunner/client/lienzo/canvas/controls/LineSpliceAcceptorControlImplTest.java
```diff
@@ -0,0 +1,523 @@
...
+
+        verify(highlight, times(2)).unhighLight();
+        verify(highlight, times(1)).highLight(connector);
+        verify(commandManager, times(1))
+                .execute(eq(canvasHandler), any(CompositeCommand.class));
+
+        verify(tested).executeCommands(any(Function.class), commandsCapture.capture());
+        final CompositeCommand<AbstractCanvasHandler, CanvasViolation> commands = commandsCapture.getValue();
+        final List<Command<AbstractCanvasHandler, CanvasViolation>> commandList = commands.getCommands();
+
```
handreyrc, line 362: Hey @domhanak,
This is the point I'm checking if all the commands needed were properly added to the batch to perform the reconnetions and the containment. This test covers the most complete scenatio. 
The idea is to ensure the all the commands for each scenatio are there.
Thanks

handreyrc:
Hi there @domhanak,

I reported the issue 03, please feel free to add any information you find necessary. I investigated a bit around the parentPicker and actually we need the actual behavior for the regular containment and for docking. So in my opinion it is better to handle it as an edge case of the line splice.
**Jira**: [KOGITO-5676](https://issues.redhat.com/browse/KOGITO-5676 )

In regards with the unit tests check the class LineSpliceAcceptorControlImplTest.java,  The idea is to check if the the commands for the different scenarios are in the batch to be executed. I'm intercepting the composite command and checking if we have to expected commands inside. 

Thanks

domhanak (approved):
Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/pom.xml
```diff
@@ -104,6 +104,11 @@
       <artifactId>appformer-client-api</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>lienzo-core</artifactId>
```
romartin, line 109: mmm I am not sure why we need to add the lienzo dependency into the stunner's client api module. The idea is that lienzo is the library used for the implementation of our client side stuff, but our "client API" should be decoupled from lienzo. This way, at some point, we should be able to move to another technology for rendering our client side stuff.

I don't think this blocks this PR, but it would be nice analyzing why adding the dependency to lienzo in our API module, and trying to see if it's easy to revert this and keep the API decoupled from the implementation library (lienzo). This can be done in a separate PR, but please, try to do this analysis and let's see why the need of this depenendency here. This creates lot of coupling and I would prefer to fix it at some point.

Thanks!

lienzo-core/src/main/java/com/ait/lienzo/client/core/shape/wires/ILineSpliceAcceptor.java
```diff
@@ -0,0 +1,70 @@
...
+
+    boolean allowSplice(WiresShape shape,
+                        Point2D candidateLocation,
+                        WiresConnector connector,
+                        WiresContainer parent);
+
+    boolean acceptSplice(WiresShape shape,
+                         Point2D candidateLocation,
+                         WiresConnector connector,
+                         Point2DArray firstHalfPoints,
```
romartin, line 36: another quick Q - I see the arguments for the `acceptSplice` contain `firstHalfPoints` and `secondHalfPoints`, but is this really necessary? I mean:

- the points array can be obtained from the `connector` instance, which is alrady an argument
- this way, buy just adding an `integer` argument, which specificies the index where to split the above connector's points, we can extrapolate same info
- so given the `connector` and some `index` (int), the `firstHalfPoints` and `secondHalfPoints` can be extrapolated

Does this makes sense? Or is not so easy? :)

Again, this not blocks this PR, just brainstorming,

romartin (approved):
Hey @handreyrc 

Really great stuff! Code looks great as well. Thanks for all this work :+1: 

Approving, although just added a few of minor comments, could you please check? If necessary, go ahead and create another small PR that fixes those comments. At least I would like to decouple the client-api module from lienzo, if possible.

Anyway, all good and ready to merge. Let's keep in touch about those comments.

Thanks!

PR #87, opened 2021-08-02
KOGITO-5469 : Can't connect elements vertically far from each other in a BPMN Diagram

inodeman:
Hi @romartin can you review this PR? Added both the fix and unit test, thanks

### DIFF
--- a/lienzo-core/src/main/java/com/ait/lienzo/client/core/shape/wires/proxy/WiresDragProxy.java
+++ b/lienzo-core/src/main/java/com/ait/lienzo/client/core/shape/wires/proxy/WiresDragProxy.java
@@ -72,8 +72,14 @@ public class WiresDragProxy {
 
     private void move(final double x,
                       final double y) {
-        final double dx = x - startPoint.getX();
-        final double dy = y - startPoint.getY();
+
+        final Transform transform = getLayer().getViewport().getTransform();
+        final Point2D inversed = new Point2D(0, 0);
+        final Point2D point = new Point2D(startPoint.getX(), startPoint.getY());
+        transform.getInverse().transform(point, inversed);
+
+        final double dx = x - inversed.getX();
+        final double dy = y - inversed.getY();
 
         getDelegate().move(dx, dy);
         proxyDragLayer.moveToTop();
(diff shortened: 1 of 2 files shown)


domhanak (changes requested):
Checked the behavior that was specified in ticket.

I was able to connect nodes far from each other on canvas. :heavy_check_mark: 
I was able to zoom the diagram out and connect the node from bottom to the node on top. :heavy_check_mark: 

However I was not able to reach the **Expected Results** specified in ticket.

Expected result

* It's possible to connect two nodes placed anywhere in the process.
* Details:
    * If the target node is not visible in the process, the displayed process area is continuously adjusting according to mouse pointer (the view is scrolled in direction of pointer movement) while the sequence flow's arrow is selected. :heavy_multiplication_x: 

This does not work, when I drag the sequence flow to the direction of node the **visible process area** is not scrolled,  focus is lost, and dragged sequence flow target is the point where the pointer left canvas.

Another thing is that when I use scroll wheel while dragging a node the focus is also lost and the dragged sequence is lost. Reported here, but I think it is a valid case for this issue.

inodeman:
Hi @domhanak discussed with @romartin. The main issue on the ticket was that when Zooming in and out the sequence flow or node was not placed correctly following the mouse pointer. I agree that the scrolling seems to be an issue, but seems that would belong to a separate ticket. I have gone ahead and updated the ticket to correctly describe the main Zoom In/Out issue as well as acceptance criteria. Let me know

domhanak:
@inodeman I disagree, the main issue was properly described in the ticket before, now it is not clear what is going to be fixes. I acked the ticket based on previous description, now I do not understand what is meant by the updated description.
Please provide a reproducer for the updated description.

domhanak:
@kmacedovarela Hi, as you are a reporter for https://issues.redhat.com/browse/RHPAM-3534 could you please check the updated description and let @romartin and @inodeman  know if you agree with the updated scope?

kmacedovarela:
Thanks for the heads up @domhanak.

@romartin @inodeman, the new description is aligned with the original reported behavior. 
Thanks!

domhanak:
In that case, we should be good to merge. Thank you :)

domhanak (approved):

domhanak:
I will check if we have JIRA for following behavior and  create a new one if not.
`If the target node is not visible in the process, the displayed process area is continuously adjusting according to mouse pointer (the view is scrolled in direction of pointer movement) while the sequence flow's arrow is selected.`

romartin (approved):

romartin:
Thanks @domhanak 

BTW I don't think we have the JIRA for your comment above, please go ahead, report it  and let's discuss it. :+1:

PR #88, opened 2021-08-03
KOGITO-5481: Add 50x50 Relation cypress test

jomarko:
Dependent on a https://github.com/kiegroup/kogito-editors-java/pull/82

### DIFF
--- a/kogito-editors-js/packages/boxed-expression-component/showcase/cypress/integration/relation_expression_spec.ts
+++ b/kogito-editors-js/packages/boxed-expression-component/showcase/cypress/integration/relation_expression_spec.ts
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// <reference types="Cypress" />
+
+describe("Relation Expression Tests", () => {
+  beforeEach(() => {
+    cy.visit("/");
+  });
+
+  it("Define 50x50 Relation expression", () => {
+    // Entry point for each new expression
+    cy.ouiaId("expression-container").click();
+
+    // Define new expression as Relation
+    cy.ouiaId("expression-popover-menu").contains("Relation").click({ force: true });
+
+    // Define 50x50 Relation
+    cy.ouiaId("edit-expression-json").click();
+
+    const relationColumns = new Array(50);
+    for (let index = 0; index < 50; index++) {
+      relationColumns[index] = { name: `column-${index}`, dataType: "<Undefined>", width: 150 };
+    }
+
+    const relationRows = new Array(50);
+    for (let rowIndex = 0; rowIndex < 50; rowIndex++) {
+      const row = new Array(50);
+      for (let columnIndex = 0; columnIndex < 50; columnIndex++) {
+        row[columnIndex] = `row ${rowIndex} column ${columnIndex}`;
+      }
+      relationRows[rowIndex] = row;
+    }
+
+    const bigRelation = {
+      name: "Expression Name",
+      dataType: "<Undefined>",
+      uid: "id1",
+      isHeadless: false,
+      logicType: "Relation",
+      columns: relationColumns,
+      rows: relationRows,
+    };
+
+    cy.ouiaId("typed-expression-json").clear();
+    cy.ouiaId("typed-expression-json").invoke("val", JSON.stringify(bigRelation)).type(" ");
+
+    cy.ouiaId("confirm-expression-json").click();
+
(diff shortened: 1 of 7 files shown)


vpellegrino (approved):
Thanks for this test case!

kogito-editors-js/packages/boxed-expression-component/showcase/cypress/integration/relation_expression_spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  beforeEach(() => {
+    cy.visit("/");
+  });
+
+  it("Define 50x50 Relation expression", () => {
+    // Entry point for each new expression
+    cy.ouiaId("expression-container").click();
+
+    // Define new expression as Relation
+    cy.ouiaId("expression-popover-menu").contains("Relation").click({ force: true });
```
jstastny-cz, line 29: The `force:true` might hide issues like element not visible, etc. Sometimes can help sth like `.contains("Relation").scrollIntoView().click()`, but depends case by case. There are cases when `force:true` helps for no obvious reason, so just checking.

kogito-editors-js/packages/boxed-expression-component/showcase/cypress/integration/relation_expression_spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      columns: relationColumns,
+      rows: relationRows,
+    };
+
+    cy.ouiaId("typed-expression-json").clear();
+    cy.ouiaId("typed-expression-json").invoke("val", JSON.stringify(bigRelation)).type(" ");
+
+    cy.ouiaId("confirm-expression-json").click();
+
+    cy.ouiaId("OUIA-Generated-Table-2").should("contain.text", "row 49 column 49");
```
jstastny-cz, line 63: Are we able to find the Table component in code and set some ouiaId explicitly? It can be a generated one, but when we  generate it ourselves, it's still more stable than the default one.

jstastny-cz (approved):
Just minor comments, fine to go even in current shape.

kogito-editors-js/packages/boxed-expression-component/showcase/package.json
```diff
@@ -4,16 +4,17 @@
...
-    "react-scripts": "^4.0.3",
-    "customize-cra": "^1.0.0",
-    "react-app-rewired": "^2.1.8"
+    "react-scripts": "^4.0.3"
   },
   "devDependencies": {
     "@types/node": "^15.0.2",
     "@types/react": "^17.0.5",
     "@types/react-dom": "^17.0.3",
+    "cypress": "^8.1.0",
```
karreiro, line 17: May we add this line in the `kogito-editors-js/package.json` as well?

karreiro (commented):
Thank you, @jomarko. Great stuff! I've left just one inline/minor comment.

kogito-editors-js/packages/boxed-expression-component/showcase/package.json
jomarko: I think yes, however not sure, if then it will be needed in `showcase`, aren't the parent dependencies available in childs?

kogito-editors-js/packages/boxed-expression-component/showcase/cypress/integration/relation_expression_spec.ts
jomarko: I tried your proposed solution, however I am getting this error in such case.

```
CypressError: Timed out retrying after 4050ms: `cy.click()` failed because this element is not visible:

`<button class="pf-c-menu__item" tabindex="-1" type="button">...</button>`
This element `<button.pf-c-menu__item>` is not visible because its content is being clipped by one of its parent elements, which has a CSS property of overflow: `hidden`, `scroll` or `auto`
Fix this problem, or use `{force: true}` to disable error checking.


```    


The captured screen seems like below:
Relation Expression Tests -- Define 50x50 Relation expression (failed)

As I am not aware of a bug in "real" application, I prefer to keep `{force: true}`

What do you think?

kogito-editors-js/packages/boxed-expression-component/showcase/cypress/integration/relation_expression_spec.ts
jomarko: for sure I will adopt this suggestion, I like it

jomarko:
@karreiro please provide your opinion about https://github.com/kiegroup/kogito-editors-java/pull/88#discussion_r682544064. Would you like to have `cypress` dependency on a single place or both in parent and child module?

@jstastny-cz I adopted a suggestion about ouia id, unfortunately I was not able to get rid of `click({force: true})`, please let us know if you have any other solution in mind, maybe form you experience with cypress tests.

kogito-editors-js/packages/boxed-expression-component/showcase/package.json
karreiro: @jomarko I believe it would be nice to have it on both places. Here we're declaring dependencies for the `showcase` project. And `kogito-editors-js/package.json` help us to keep all dependencies centralized in a single place (so, all projects may share the same versions).

kogito-editors-js/packages/boxed-expression-component/showcase/cypress/integration/relation_expression_spec.ts
jstastny-cz: Fair enough, visually this looks correct, feel free to leave the `force:true`.

jstastny-cz:
@jomarko Looks good, thank you.

kogito-editors-js/packages/boxed-expression-component/showcase/package.json
jomarko: @karreiro changes are in

karreiro (approved):
Thank you, @jomarko!

PR #101, opened 2021-08-30
KOGITO-5149: Create the second step of the Wizard - Create the collapsible/expandable list of future Data Types

yesamer:
https://issues.redhat.com/browse/KOGITO-5149

Demo avaliable here: https://yesamer.github.io/kogito-editors-java/

### DIFF
--- a/kogito-editors-js/packages/import-java-classes-component/src/__tests__/ImportJavaClasses/ImportJavaClasses.test.tsx
+++ b/kogito-editors-js/packages/import-java-classes-component/src/__tests__/ImportJavaClasses/ImportJavaClasses.test.tsx
@@ -18,7 +18,6 @@ import * as React from "react";
 import { fireEvent, render } from "@testing-library/react";
 import { ImportJavaClasses } from "../../components";
 import * as _ from "lodash";
-import { act } from "react-dom/test-utils";
 
 describe("ImportJavaClasses component tests", () => {
   test("should render ImportJavaClasses Button component", () => {
@@ -77,11 +76,11 @@ describe("ImportJavaClasses component tests", () => {
     checkFirstElement = baseElement.querySelector('[aria-labelledby="com.Book"]')! as HTMLInputElement;
     expect(checkFirstElement).toBeChecked();
     expect(checkSecondElement).not.toBeChecked();
-    await fireEvent.click(checkSecondElement);
+    fireEvent.click(checkSecondElement);
     checkSecondElement = baseElement.querySelector('[aria-labelledby="com.Author"]')! as HTMLInputElement;
     expect(checkFirstElement).toBeChecked();
     expect(checkSecondElement).toBeChecked();
-    await fireEvent.click(checkSecondElement);
+    fireEvent.click(checkSecondElement);
     checkSecondElement = baseElement.querySelector('[aria-labelledby="com.Author"]')! as HTMLInputElement;
     expect(checkFirstElement).toBeChecked();
     expect(checkSecondElement).not.toBeChecked();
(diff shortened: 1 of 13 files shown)


karreiro:
@yesamer Could you please update the Demo URL? :-)

karreiro (commented):
Thank you for this PR, @yesamer!

I believe we may not covering the fetch-fields mechanism. I will try to clarify some steps from [karreiro.com/wiki/dmn/import-java-classes](https://karreiro.com/wiki/dmn/import-java-classes/#2-cherry-pick-fields) (but, the whole picture still lives there) ;)

1) Notice that the user selected these classes: **Person**, **Address**, and **UUID**:



2) So, in the second step, we notice that the fields **address** and **uuid** from **Person** already have a custom data type. Also, notice that the type of the **country** field (from **Address**) could not be inferred, so its type is **Any**:



3) Still, in the second step, the user may fetch **Country** fields by clicking in the **Fetch Country class** button, and this is the result:



Thanks again for this PR, @yesamer. And, please, let me know if need any extra clarification abou the fetch-fields mechanism :)

yesamer:
@karreiro Thank you for explanation, I'm aware of the Fetch mechanism you designed, my intention is to implement it in the next ticket: https://issues.redhat.com/browse/KOGITO-5150.
The scope of this ticket is to introduce the second step show the types only.

karreiro (commented):
Thanks a lot for the clarification, @yesamer! I have only one minor last comment before the approval 😅

At this point, Java classes are already being handled as DMN data types:
Screenshot 2021-09-02 at 13 26 15

I believe it would be more clear with we remove the `(org.kie.test.kogito.Book)`, or with we rename it as `(Structure)`. Wdyt? (also, please let me know if you prefer to keep it as is).

Thanks again for this PR! :-)

kogito-editors-js/packages/import-java-classes-component/src/__tests__/ImportJavaClasses/ImportJavaClasses.test.tsx
```diff
@@ -99,9 +98,83 @@ describe("ImportJavaClasses component tests", () => {
     expect(baseElement).toMatchSnapshot();
   });
 
+  test("Should move to second step", async () => {
```
vpellegrino, line 101: It seems we are awaiting anything. Is this `async` really needed?
Also, a suggestion from a readability perspective: maybe you could split the code of this test in meaningful methods

vpellegrino (approved):
PR is fine from my point of view!
Just minor feedback on the test case.

Thanks, @yesamer

domhanak:
Not sure that I see the latest demo on the URL. I can only see Book and Boom data types and can't use org.* for example in search. Please clarify.
Anyway i checked the demo and it worked well, although I was not able to select the fields ( no option to select them) Is that expected? The JIRA has no description on whats supposed to be deemed as Acceptable so I really do not know what to actually verify :D

yesamer:
> Not sure that I see the latest demo on the URL. I can only see Book and Boom data types and can't use org.* for example in search. Please clarify.
> Anyway i checked the demo and it worked well, although I was not able to select the fields ( no option to select them) Is that expected? The JIRA has no description on whats supposed to be deemed as Acceptable so I really do not know what to actually verify :D

@domhanak
 You're right, I'll give you more context in the ticket description. I'll ping you on Monday. Thx!

kogito-editors-js/packages/import-java-classes-component/src/__tests__/ImportJavaClasses/ImportJavaClasses.test.tsx
yesamer: @vpellegrino I applied your suggestions. Thx u!

karreiro (approved):
Thank you, @yesamer! (approving, as we will handle the type inference in another PR) :)

yesamer:
@domhanak 


> Not sure that I see the latest demo on the URL. I can only see Book and Boom data types and can't use org.* for example in search. Please clarify.

Currently, the service that retrieves the data from Java Classes is totally mocked. It works only if the user types `Boo`,  `Boom` or `Book`. To enable this feature we need the LSP server framework, which is currently a WIP.

> Anyway i checked the demo and it worked well, although I was not able to select the fields ( no option to select them) Is that expected? The JIRA has no description on whats supposed to be deemed as Acceptable so I really do not know what to actually verify :D
I provided a description in the JIRA, basically, for this PR no actions are planned in the second step. In another ticket (5150), I'll will introduce the `Fetch` action mechanism (I'll provide more info in the 5150 ticket). If you need futher info, please feel free to ask here or in the JIRA ticket or ping me in chat :)

domhanak (approved):
Thanks for clarification. LGTM

PR #102, opened 2021-09-06
WIP KOGITO-5468: DMN Tabular Input

ljmotta:
## WIP PR

On this PR:
- Modify the boxed-component to enable dynamic data as input so it can be used by the DMN Runner.
- Add a new build that will be used to publish the component on npm.

## Known issues
- All recursive components (Invocation, Context, Function, List) returns to their initial state after an update is made on a field that has being already assign to a field type. 
e.g. Create a Context field, change the first row to be a "Literal Expression". Add new rows, updating the first literal expression will erase all previously add rows.
- Function PMML: second Select field can't be changed.


tiagobento:
@karreiro Hi! This is a WIP PR, however, I think we would benefit a lot if you good take a look at the changes Luiz already made. We needed to change places not directly related to our use-case to enable it, and it would be great if you could take a look and provide us with feedback to know if we're going on the right direction. Thanks in advance!!

karreiro:
Thank you for this PR, @ljmotta - also, thanks for your thoughts, @tiagobento.

I'll review it for sure, but @vpellegrino is the SME for the new Boxed Expression component. I believe he can provide valuable thoughts from the architecture perspective (essentially from the update-propagation-mechanism aspect).

Thanks again for this effort! 🚀

tiagobento:
@vpellegrino Hi Valentino! Please reach out if you have questions about something on the code!

vpellegrino:
Hi @ljmotta and @tiagobento 
I will start to review this PR today. Maybe it will require some time, but I will reach out to you immediately if anything is unclear.
Thanks for this PR :)

vpellegrino:
I've checked out your branch, firstly to understand the overall solution and related architecture.
That's a thing I've noticed: the `boxed-expression-components`'s `quality-checks` script (https://github.com/kiegroup/kogito-editors-java/blob/f0912ca114b23ddcbd466bdaf9eb8cdb1204ec51/kogito-editors-js/packages/boxed-expression-component/package.json#L18) is failing.

vpellegrino:
I've not yet started to look at code, file by file, but I wanted to reach out to you immediately. I have two observations:
1) Should `dmn-runner-showcase` folder be **not** inside `boxed-expression-component`? Logically it is a self-contained project that *uses* the `boxed-expression-component`. Any technical reason for not being an isolated project which has `boxed-expression-component` as a dependency? Furthermore, I think that a `README.md` would be really appreciated, to understand the project's purpose is and how to compile&launch it.
2) Typically, when I need to use an external component, and I need it to behave slightly differently to cover my use case, I prefer to wrap it with another component to avoid modifying it. I can see that in this specific case, the `boxed-expression-component` has been modified a lot. I don't have any specific opposition on that, but it is important to ensure that the existing component's use cases are still covered and that no regression has been introduced. I just launched `boxed-expression-component`'s showcase, and I noticed that:
- `<result>` row is missing for the context expression:
image
- it is not possible to edit a `Relation` table with multiple cells:
image

If we want to proceed in this direction, a QE member must help us test it, for example, @jomarko. 
Furthermore, consider that we are working to integrate the `boxed-expression-component` inside the existing `DMN Editor` (https://github.com/kiegroup/kogito-editors-java/pull/104), and we need to make sure that the changes to it do not affect the integration itself.

tiagobento:
Hi @vpellegrino, thanks for taking a look at this PR! I appreciate you taking some time to thoroughly look at it and find bugs, as Luiz and I are not familiar with all the use cases it has.

Of course, wrapping the existing component and adding features on top of it is the ideal scenario and is what we were aiming for, but it wasn't possible, so we had to change it internally. This PR is the beginning of a big effort to make the same "table" component be used in both the new Boxed Expressions and the DMN Runner Tabular Input.

We tried to keep changes to a minimum to avoid breaking existing use-cases, but it's a complex task, so naturally bugs appeared, as you found out. We're going to keep working on this PR until we have no regressions and we're comfortable with the DMN Runner Tabular Input quality.

Packaging and bundling is *not* final, as with most things on this PR. We're also going to have big changes soon, as this repository is going to be moved to `kogito-tooling` as part of our Monorepo initiative, so don't look to much into those areas just yet 😝 

There is a lot to figure out, but ultimately we're going to have a unique experience on all of our table-formatted data. Looking forward to it!

Thanks again for taking a preliminary look at this PR, we'll get back to it as soon as Luiz is back from PTO :)

jomarko (changes requested):
### Context
- as said, last result row is missing

### Decision Table
- insert right doesn't work at all
- insert left causes some infinite loop (can be seen on the horizontal scrollbar)
- repeated start edit (by enter), navigating among cells by shift + tab/tab, causes some cells starts to be selected and text to disappear, that is general issue for all expression types

### Function
- We need to disable insert context menu action for java and pmml functions. for java only class name and signature is allowed. For pmml only document and mode;.

### showcase/cypress tests
are broken by changes introduced by this PR

### dmn-runner-showcase
missing instructions to build and start, any info missing about this module

ljmotta:
@jomarko @vpellegrino  I've created a new PR https://github.com/kiegroup/kogito-editors-java/pull/145 due to the size of the necessary modifications to solve all issues. I'll be closing this one.

PR #104, opened 2021-09-08
KOGITO-3661 Integrate boxed-expression-component inside dmn-loader and define jsInterop classes for loading/saving expressions

vpellegrino:
Showcase `boxed-expression-component`: https://boxed-expression-editor.netlify.app/
Tasks solved by this PR: KOGITO-3666, KOGITO-3663, KOGITO-3661

Since the goal of integrating the `boxed-expression-component` inside the `DMN Editor` is ambitious, the process will be incremental, thanks to the usage of _Stable/Beta_ flags.


jomarko:
### 01 - Manual Review - Custom Type and Expression Clear
If the Decision node has a custom data type set, this is lost after expression is cleared, see the attached video https://issues.redhat.com/secure/attachment/12641782/custom-type-expression-clear.webm

jomarko:
### 02 - Manual Review - Column width
After reopen column widh is different

Before
Screenshot from 2021-09-13 13-58-33

After
Screenshot from 2021-09-13 13-58-58

jomarko:
### 03 Manual Review - Default column width
Column width seems to be different for columns created after expression definition vs. new column added.

Compare **column-2** and **annotation-2** which were added after expression was defined.

Screenshot from 2021-09-13 14-04-41

jomarko:
### 04 Manual Review - Invocation reopen

Invocation can not be reopened with new editor.
Screenshot from 2021-09-13 14-17-40

https://issues.redhat.com/secure/attachment/12641785/invocation.dmn

jomarko:
### 04 Manual Review - Nested Relation
Nested relation causes expression to disappear.

Screenshot from 2021-09-13 14-27-55


Screenshot from 2021-09-13 14-28-35

jomarko:
### 06 Manual Review - Scroll
Context expression with many entries can not be scrolled down

Screenshot from 2021-09-13 14-50-00

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
```diff
@@ -294,6 +434,79 @@ private void preventDefault(final ClickEvent event) {
...
+        return hasName.orElse(fallbackHasName).getValue().getValue();
+    }
+
+    @SuppressWarnings("unchecked")
+    private String getTypeRef() {
+        QName qName = BuiltInType.UNDEFINED.asQName();
+        if (hasExpression instanceof HasVariable) {
+            qName = ((HasVariable<InformationItemPrimary>) hasExpression).getVariable().getTypeRef();
+        } else if (hasExpression.getExpression() != null && hasExpression.getExpression().asDMNModelInstrumentedBase().getParent() instanceof HasVariable) {
+            qName = ((HasVariable<InformationItemPrimary>) hasExpression.getExpression().asDMNModelInstrumentedBase().getParent()).getVariable().getTypeRef();
```
jomarko, line 448: are we sure `getVariable()` is always `!= null`?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
```diff
@@ -294,6 +434,79 @@ private void preventDefault(final ClickEvent event) {
...
+    @SuppressWarnings("unchecked")
+    private void setTypeRef(final String dataType) {
+        final QName typeRef = BuiltInTypeUtils
+                .findBuiltInTypeByName(dataType)
+                .orElse(BuiltInType.UNDEFINED)
+                .asQName();
+        if (hasExpression instanceof HasVariable) {
+            ((HasVariable<InformationItemPrimary>) hasExpression).getVariable().setTypeRef(typeRef);
+        } else if (hasExpression.getExpression() != null && hasExpression.getExpression().asDMNModelInstrumentedBase().getParent() instanceof HasVariable) {
+            ((HasVariable<InformationItemPrimary>) hasExpression.getExpression().asDMNModelInstrumentedBase().getParent()).getVariable().setTypeRef(typeRef);
```
jomarko, line 472: similar question here

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/jsinterop/util/ExpressionFiller.java
```diff
@@ -0,0 +1,596 @@
...
+            final DecisionTable decisionTableExpression = (DecisionTable) wrappedExpression;
+            final String hitPolicy = decisionTableExpression.getHitPolicy() != null ? decisionTableExpression.getHitPolicy().value() : null;
+            final String aggregation = decisionTableExpression.getAggregation() != null ? decisionTableExpression.getAggregation().getCode() : "";
+            return new DecisionTableProps(expressionName, dataType, hitPolicy, aggregation,
+                                          annotationsConvertForDecisionTableProps(decisionTableExpression),
+                                          inputConvertForDecisionTableProps(decisionTableExpression),
+                                          outputConvertForDecisionTableProps(decisionTableExpression),
+                                          rulesConvertForDecisionTableProps(decisionTableExpression));
+        }
+        return new ExpressionProps(expressionName, dataType, null);
```
jomarko, line 189: Shouldn't we throw some exception about illegal state?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/jsinterop/util/ExpressionFiller.java
```diff
@@ -0,0 +1,596 @@
...
+        } else if (FUNCTION.getText().equals(props.logicType)) {
+            final FunctionDefinition functionExpression = new FunctionDefinition();
+            fillFunctionExpression(functionExpression, (FunctionProps) props);
+            return functionExpression;
+        } else if (DECISION_TABLE.getText().equals(props.logicType)) {
+            final DecisionTable decisionTableExpression = new DecisionTable();
+            fillDecisionTableExpression(decisionTableExpression, (DecisionTableProps) props);
+            return decisionTableExpression;
+        }
+        return null;
```
jomarko, line 226: Shouldn't we throw some exception about illegal state?

jomarko (commented):
`ExpressionFiller` seems to contain a the main part of the logic related to changes introduced by this PR, shouldn't we test this class?

Is this class just a  temporary?

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/jsinterop/props/DecisionTableProps.java
```diff
@@ -0,0 +1,48 @@
...
+    public final String aggregation;
+    public final Annotation[] annotations;
+    public final Clause[] input;
+    public final Clause[] output;
+    public final DecisionTableRule[] rules;
+
+    public DecisionTableProps(final String name,
+                              final String dataType,
+                              final String hitPolicy,
+                              final String aggregation,
```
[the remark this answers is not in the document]
vpellegrino: In the model class, we have two separate fields, so we are doing the same also for the props class.

kogito-editors-js/packages/boxed-expression-component/src/components/Resizer/common/index.ts
```diff
@@ -27,7 +27,20 @@ export const widthValue = (width: number | string | undefined | null): number =>
...
+        if (seen.has(value)) {
+          return;
+        }
+        seen.add(value);
+      }
+      return value;
+    };
+  };
+
+  return JSON.stringify(obj, getCircularReplacer());
```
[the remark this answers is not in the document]
vpellegrino: Since `jsInterop` layer adds circular references inside Java objects that are translated into JS Objects, the `stringify` method would fail with such circular references. There is a proper solution provided by MDN (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value#examples)

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
vpellegrino: Maybe I'm misleading the type check here. We are checking if `getParent()` is instance of `HasVariable`. Isn't this enough to be sure that `getVariable` is always `!= null`? I will add an extra check.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
vpellegrino: https://github.com/kiegroup/kogito-editors-java/pull/104#discussion_r707979290

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/jsinterop/util/ExpressionFiller.java
vpellegrino: I don't think. It is the case when an empty expression is translated to JS props.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/jsinterop/util/ExpressionFiller.java
vpellegrino: In that case, we are filling an empty nested expression model. I've tried, and it seems that when the nested expression is not set, we should set `null` for it. So, I would say it is a normal flow that an exception should not block.

vpellegrino:
> `ExpressionFiller` seems to contain a the main part of the logic related to changes introduced by this PR, shouldn't we test this class?
> 
> Is this class just a temporary?

Thanks for your valuable feedback @jomarko!
I will work to introduce a unit test for the `ExpressionFiller` class and work also on all other issues you found (thanks a lot for that).

danielzhe:
@vpellegrino For you info: about the tests, if you are planning to cover more than `ExpressionFiller` as @jomarko pointed, don't worry too much about `ExpressionEditorViewImpl` because it will be changed with the undo/redo. Most of its logic will be in commands.

vpellegrino:
> @vpellegrino For you info: about the tests, if you are planning to cover more than `ExpressionFiller` as @jomarko pointed, don't worry too much about `ExpressionEditorViewImpl` because it will be changed with the undo/redo. Most of its logic will be in commands.

Ok @danielzhe thanks for your feedback!
I'm going to add tests to cover `ExpressionFiller`

vpellegrino:
@jomarko, let me update you:
01) Https://issues.redhat.com/browse/KOGITO-4588 will cover it - feel free to mention this acceptance criterion
02) It Seems an issue related to the `Resizer` feature. @karreiro is looking on. To unblock this PR, we may trace it with a separate JIRA. Let me know.
03) Solved
04) Solved
05) Solved
06) Solved

Covered `Filler` classes with test cases. As stated by @danielzhe, the `ExpressionEditorViewImpl` will be extended and covered by his test cases.

Thanks again for your review. Would you mind letting me know if you have any further doubt?

jomarko:
@vpellegrino would you mind to have a look on this bug reported by sonar? https://sonarcloud.io/project/issues?id=org.kie.kogito%3Akogito-editors-java&pullRequest=104&resolved=false&types=BUG

jomarko:
### re review

1 -  I am fine with separate issue
2 - reported https://issues.redhat.com/browse/KOGITO-5909
3 - check in progress
4 - check in progress
5 - check in progress
6 - check in progress
code re review - check in progress

kie-wb-common-dmn/kie-wb-common-dmn-client/src/test/java/org/kie/workbench/common/dmn/client/editors/expressions/jsinterop/util/ExpressionModelFillerTest.java
```diff
@@ -0,0 +1,334 @@
...
+        assertThat(componentWidths).element(1).isEqualTo(ENTRY_INFO_WIDTH);
+        assertThat(componentWidths).element(2).isEqualTo(ENTRY_EXPRESSION_WIDTH);
+    }
+
+    private void assertFormalParameters(FunctionDefinition functionExpression) {
+        assertThat(functionExpression.getFormalParameter()).isNotNull();
+        assertThat(functionExpression.getFormalParameter()).hasSize(1);
+        assertThat(functionExpression.getFormalParameter()).first().extracting(param -> param.getValue().getValue()).isEqualTo(PARAM_NAME);
+        assertThat(functionExpression.getFormalParameter()).first().extracting(param -> param.getTypeRef().getLocalPart()).isEqualTo(PARAM_DATA_TYPE);
+    }
```
jomarko, line 316: Generally I think, we overuse `assertThat` what cause some assertions less readable, one example how come can be simplified:
```
    private void assertFormalParameters(FunctionDefinition functionExpression) {
        assertThat(functionExpression.getFormalParameter())
                .isNotNull()
                .hasSize(1)
                .first()
                .satisfies(param -> {
                    assertThat(param.getValue().getValue()).isEqualTo(PARAM_NAME);
                    assertThat(param.getTypeRef().getLocalPart()).isEqualTo(PARAM_DATA_TYPE);
                });
    }
```

jomarko (commented):
Some comments to code, no more defects found in manual review. Thank you @vpellegrino

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/jsinterop/util/ExpressionModelFiller.java
```diff
@@ -0,0 +1,402 @@
...
+        return Arrays
+                .stream(Optional.ofNullable(relationProps.rows).orElse(new String[0][]))
+                .map(row -> {
+                    final List list = new List();
+                    list.getExpression().addAll(
+                            IntStream.range(0, Optional.ofNullable(relationProps.columns).orElse(new Column[0]).length).mapToObj(columnIndex -> {
+                                final String cell = row.length <= columnIndex ? "" : row[columnIndex];
+                                final LiteralExpression wrappedExpression = new LiteralExpression();
+                                wrappedExpression.setText(new Text(cell));
+                                wrappedExpression.setTypeRef(BuiltInType.STRING.asQName());
```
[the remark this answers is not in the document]
vpellegrino: In this method, we are setting the `typeRef` of the wrapped `LiteralExpression` (used for the cell) to `string`. I'm going to remove it. Thanks for letting me know

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/jsinterop/util/ExpressionModelFiller.java
```diff
@@ -0,0 +1,402 @@
...
+    private static Collection<InformationItem> columnsConvertForRelationExpression(final RelationProps relationProps) {
+        return Arrays
+                .stream(Optional.ofNullable(relationProps.columns).orElse(new Column[0]))
+                .map(column -> {
+                    final InformationItem informationItem = new InformationItem();
+                    informationItem.setName(new Name(column.name));
+                    informationItem.setTypeRef(BuiltInTypeUtils
+                                                       .findBuiltInTypeByName(column.dataType)
+                                                       .orElse(BuiltInType.UNDEFINED)
+                                                       .asQName());
```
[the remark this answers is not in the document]
vpellegrino: FYI: This method is about settings `name` and `typeRef` for columns. As per https://github.com/kiegroup/kogito-editors-java/pull/104#discussion_r710089048, I'm avoiding to set `string` as typeRef for wrapped expression

vpellegrino:
> Some comments to code, no more defects found in manual review. Thank you @vpellegrino

@jomarko answered, and accomplished comments as well.
Thank you!

jomarko (approved):
@vpellegrino thank you

danielzhe:
Hi, @vpellegrino ! I've been working with this branch to do undo/redo. Very good job!
I found just a strange issue. Looks like that when we set a `Function` we got a infinite call to `broadcastFunctionExpressionDefinition(final FunctionProps functionProps)` https://github.com/kiegroup/kogito-editors-java/pull/104/files#diff-165ca4e61d69efdf36647961ef3e2a1d315633ada4516c6137f7b7c4959f3d21R404

vpellegrino:
> Hi, @vpellegrino ! I've been working with this branch to do undo/redo. Very good job!
> I found just a strange issue. Looks like that when we set a `Function` we got a infinite call to `broadcastFunctionExpressionDefinition(final FunctionProps functionProps)` https://github.com/kiegroup/kogito-editors-java/pull/104/files#diff-165ca4e61d69efdf36647961ef3e2a1d315633ada4516c6137f7b7c4959f3d21R404
> 
> To reproduce, just put a break point there and set the expression to "Function". You'll notice that it starts to loop forever.

Hi @danielzhe yeah, I already noticed it.
Indeed, it is an issue caused by the Resizer (the class responsible for auto resize the tables) that @karreiro  is already working on!
Thanks a lot for raising the flag. It's nice you help in debugging the editor 😉☺️

karreiro (commented):
Thank you for this PR, @vpellegrino. Great stuff!

I have only two questions:

#### 1) Enable it on Kogito channels

I've noticed we're removing the `display: none` property from the `ExpressionEditorViewImpl.html` template. So, we're actually enabling this component in the editor.


However, only the `kie-wb-common-dmn-webapp-kogito-testing` web app requires our React assets. Thus, I believe if we build the `.vsix` extension or other Kogito channel, our component won't load (because Kogito channels rely on the `kie-wb-common-dmn-webapp-kogito-runtime` module).

Thus, I believe we have two alternatives:
- Add the static assets to the `the kie-wb-common-dmn-webapp-kogito-runtime`
- Bring the `"display: none"` back to the `ExpressionEditorViewImpl.html` and enable the component in another PR

Wdyt about this?

#### 2) Use of JSNI

We're relying on **JSNI** in the `BoxedExpressionService.java,` which is something we must really avoid; instead, we should rely on the **JsInterop** API like we're doing in the `DMNLoader.java`. Do you prefer to raise a Jira to handle this in a different PR or handle it here?

--

Thanks a lot for this PR, @vpellegrino! It's really exciting to see the new boxed expression editor alive 🚀

vpellegrino:
@karreiro 
1) I put the `"display: none"` back
2) should be fixed

karreiro (approved):
Thank you, @vpellegrino 🚀

PR #110, opened 2021-09-14
Fix FDB build

tiagobento:


.github/workflows/ci_full_downstream_build.yml
```diff
@@ -115,138 +102,37 @@ jobs:
...
+          KOGITO_TOOLING_BUILD_test: "true"
+          KOGITO_TOOLING_BUILD_testIT: "true"
+          KOGITO_TOOLING_BUILD_docker: "true"
+          KOGITO_TOOLING_BUILD_examples: "true"
           DISPLAY: ":99.0"
+          START_SERVER_AND_TEST_INSECURE: "true"
         run: |
-          cd ${{ github.workspace }}/kogito-tooling
-          yarn bootstrap && yarn build:prod
+          lerna run build:prod --stream --concurrency 1
```
domhanak, line 123: so no `yarn bootstrap` before we begin building? Is that expected?

tiagobento: It's part of the `setup-env` action on `kogito-tooling`.

tiagobento:
Merging as we need to do it to test.

.github/workflows/ci_full_downstream_build.yml
domhanak: How is that action run on `kogito-editors-java` ? if it is a action of `kogito-tooling` am I missing something?

domhanak: But imho this part will be missing `cd ${{ github.workspace }}/kogito-tooling` the error in #114 suggests that

PR #116, opened 2021-09-20
KOGITO-5909: [DMN Designer] New Boxed Expression editor - DMN Expression column width is updated too many times in some scenarios

karreiro:
Jira: [KOGITO-5909](https://issues.redhat.com/browse/KOGITO-5909): [DMN Designer] New Boxed Expression editor - DMN Expression column width is updated too many times in some scenarios

**Before:**

**After:**

### DIFF
--- a/kogito-editors-js/packages/boxed-expression-component/src/__tests__/components/Resizer/dom/Cell.test.tsx
+++ b/kogito-editors-js/packages/boxed-expression-component/src/__tests__/components/Resizer/dom/Cell.test.tsx
@@ -28,6 +28,12 @@ let cell: Cell;
 let element: HTMLElement;
 let container: Element;
 
+const EMPTY_LIST_OF_NODES = {
+  length: 0,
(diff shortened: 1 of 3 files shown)


vpellegrino (approved):
Great job, @karreiro

danielzhe:
Hi, @karreiro ! Thanks for your PR! I found out that for "Decision Table" and "Relation" the call to `broadcastRelationExpressionDefinition` is happening two times, which cause issues with undo/redo since we have duplicated commands for those cases. Can you check it, please? Thank you!

karreiro:
Thanks for the review, @danielzhe! I've not reproduced with the undo/redo feature, but I believe the issue is fixed now, see:

Before:
Kapture 2021-09-22 at 11 51 03

After:
Kapture 2021-09-22 at 11 49 12

Could you please take a look?

Thanks again 🚀  :-)

danielzhe (approved):
Great job! I didn't found any issue now! Thank you, @karreiro !

PR #119, opened 2021-09-27
JBPM-8504 - Scroll automatically expandable canvas

Josephblt:
The canvas now expands automatically anytime a primitive is dragged beyond control limits. An example was added to the web app (Infinite Canvas).

### DIFF
--- a/lienzo-core/src/main/java/com/ait/lienzo/client/widget/panel/impl/LienzoPanelHandlerManager.java
+++ b/lienzo-core/src/main/java/com/ait/lienzo/client/widget/panel/impl/LienzoPanelHandlerManager.java
@@ -186,6 +186,16 @@ public final class LienzoPanelHandlerManager {
 
         nodeDragStartEvent = new NodeDragStartEvent(m_lienzoElm);
 
+        m_lienzoElm.addEventListener("pointerdown", e ->
+        {
+            setPointerCapture(m_lienzoElm, e);
+        });
+
+        m_lienzoElm.addEventListener("pointerup", e ->
+        {
+            releasePointerCapture(m_lienzoElm, e);
+        });
+
         addEventListener(EventType.CLICKED, (Event event) ->
         {
             MouseEvent mouseEvent = (MouseEvent) event;
@@ -547,6 +557,8 @@ public final class LienzoPanelHandlerManager {
 
             fireEvent(mouseEvent, touchEvent, x, y, m_dragContext, m_drag_node.asNode(), nodeDragEndEvent);
 
+            LienzoPanelEvents.firePrimitiveDragEndEvent(m_lienzo, m_drag_node);
+
             m_dragContext.dragDone();
 
             m_drag_node.setDragging(false);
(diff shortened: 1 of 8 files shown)


romartin:
Hey @Josephblt 

The issue JBPM-8504 is not related to stunner, it's another topic and already closed. Your commits also refer to this jira. Could you please update PR's description , and commits message if necessary, please?

lienzo-core/src/main/java/com/ait/lienzo/client/widget/panel/impl/LienzoPanelHandlerManager.java
```diff
@@ -852,4 +867,12 @@ private void checkPressedMouseButton(final int nativeButtonCode) {
         m_mouse_button_middle = nativeButtonCode == MouseEventUtil.BUTTON_MIDDLE;
         m_mouse_button_right = nativeButtonCode == MouseEventUtil.BUTTON_RIGHT;
     }
+
+    private static native void setPointerCapture(HTMLElement element, Event event) /*-{
```
romartin, line 871: jsni to j2cl

lienzo-core/src/main/java/com/ait/lienzo/client/widget/panel/impl/LienzoPanelHandlerManager.java
```diff
@@ -852,4 +867,12 @@ private void checkPressedMouseButton(final int nativeButtonCode) {
...
+        element.setPointerCapture(event.pointerId)
+    }-*/;
+
+    private static native void releasePointerCapture(HTMLElement element, Event event) /*-{
```
romartin, line 875: jsni to j2cl

lienzo-core/src/main/java/com/ait/lienzo/client/widget/panel/impl/ScrollablePanel.java
```diff
@@ -299,6 +330,107 @@ private void onScroll() {
...
+        if (primMinX - padding < visMinX) {
+            dragDifferenceX = visMinX - primMinX;
+        } else if (primMaxX + padding > visMaxX) {
+            dragDifferenceX = visMaxX - primMaxX;
+        }
+    }
+
+    private void onPrimitiveDragUpdate(IPrimitive<?> primitive) {
+        final BoundingBox primitiveBoundingBox = new BoundingBox();
+        for (Point2D point2D : primitive.getComputedBoundingPoints()) {
```
romartin, line 364: Calculating BoundingBox'es or BondingPoints is expensive, and this call here is done on each drag update (for every pixel), so thsi method will be called several times when just dragging some primitive shape a little bit.Could be possible to avoid it in some way? 
If you just need the new primitive location (while dragging), the idea is to keep the inital x / y on dragStart, and then every dragUpdate call just need to increase/decrease both x/y properly, this is the way lienzo core works and avoids even accessing the primitive shape in memoty to obtain location, this is safe and fast, you can looks for examples in classes in lienzo.

romartin (commented):
Hey @Josephblt 

Thanks for this work, was not easy and I think you got a nice solution, altough I would like to check a bit more in detail the code. For now I just added some review comments, please take a look.

Thanks!

PR #127, opened 2021-10-01
KOGITO-6021: [DMN Designer] Multiple DRDs - Renaming a DRD freezes the browser

yesamer:
JIRA: https://issues.redhat.com/browse/KOGITO-6021

Steps to reproduce: 
- Create a Decision Node;
- Click on DRD Action -> Create
- Try to update the `new-diagram` that appeared above (Press ENTER when the new value is assigned)
- Browser freezes

Fix: During every rename action, a process is fired to update the reference in multiple places of the editor.  This process is fired when the user press ENTER in the input element OR the input element loses focus (blur) (eg. pressing in another point of the page).
Unfortunately, pressing ENTER will start the process and at its end, a blur event is fired duplicating the process which leads eventually to freeze the browser.
To fix it, I start the rename update process ONLY if the new value is different than the original.

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/drd/DRDNameChangerView.java
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/drd/DRDNameChangerView.java
@@ -137,7 +137,9 @@ public class DRDNameChangerView implements DRDNameChanger {
     }
 
     private void performSave(final DMNDiagramElement dmnDiagramElement) {
-        dmnDiagramElement.getName().setValue(drdNameInput.getValue());
-        selectedEvent.fire(new DMNDiagramSelected(dmnDiagramElement));
+        if (!dmnDiagramElement.getName().getValue().equals(drdNameInput.getValue())) {
+            dmnDiagramElement.getName().setValue(drdNameInput.getValue());
+            selectedEvent.fire(new DMNDiagramSelected(dmnDiagramElement));
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/drd/DRDNameChangerView.java
```diff
@@ -137,7 +135,16 @@ void saveForTheCurrentDiagram() {
     }
 
     private void performSave(final DMNDiagramElement dmnDiagramElement) {
-        dmnDiagramElement.getName().setValue(drdNameInput.getValue());
-        selectedEvent.fire(new DMNDiagramSelected(dmnDiagramElement));
+        if (!dmnDiagramElement.getName().getValue().equals(drdNameInput.getValue())) {
```
jomarko, line 138: It is a minor thing, but I prefer `Objects.equals(a, b)` rather than `a.equals(b)`. as the first mentioned shouldn't throw a `NullPointerException` in case `a` or `b` is null`.
```
if (!Objects.equals(dmnDiagramElement.getName().getValue(), drdNameInput.getValue())) {
```

jomarko (commented):
thank you for a PR @yesamer , I have one comment to a code.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/drd/DRDNameChangerView.java
yesamer: @jomarko Done!

jomarko:
@yesamer the full downstream build failed due to chromedriver incompatibility in tests. It should be fixed now https://github.com/kiegroup/kogito-tooling/commit/ed114772ca4d7d01fbff60d765ef06bf3534bdba

Could you please force-push this PR to rerun downstream build? Or do we have some message trigger we can put as github comment?

yesamer:
@jomarko Relaunched

danielzhe (approved):
Good job! Thank you, Yeser!

jomarko (approved):
Thank you @yesamer .

Just FYI, I prepared also selenium integration test for this #131

PR #146, opened 2021-10-19
KOGITO-6132: [DMN Designer] Pressing backspace when editing a node deletes the node

danielzhe:
VSIX: https://drive.google.com/file/d/1SBl1r35MtzrYNsE2wgXkioLQqd_gW9ai/view?usp=sharing
Ticket: https://issues.redhat.com/browse/KOGITO-6132

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/LienzoCanvas.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/canvas/LienzoCanvas.java
@@ -139,4 +139,9 @@ public abstract class LienzoCanvas<V extends LienzoCanvasView>
     ViewEventHandlerManager getEventHandlerManager() {
         return eventHandlerManager;
     }
+
+    @Override
+    public boolean isEventHandlesEnabled() {
+        return getEventHandlerManager().isEnabled();
+    }
 }
--- a/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/ViewEventHandlerManager.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-client/kie-wb-common-stunner-lienzo/src/main/java/org/kie/workbench/common/stunner/client/lienzo/shape/view/ViewEventHandlerManager.java
@@ -353,7 +353,7 @@ public class ViewEventHandlerManager {
         }
     }
 
-    private boolean isEnabled() {
+    public boolean isEnabled() {
         return this.enabled;
     }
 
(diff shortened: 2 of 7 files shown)


danielzhe:
Thank you, @jomarko ! Your proposed change breaks the tests (not the one that I wrote), so I think that there is a reason for that extra shortcut callback. I'm not exactly sure what is the situation, but I'm afraid that change that it can break something else, since the tests are point me to "do not do that".

romartin:
hey @jomarko @yesamer 

Sorry for late response, I just wanted to check code locally first, before saying something. 

About the `session.getKeyboardControl().addKeyShortcutCallback(getOnKeyDownEvent())` in`DeleteSelectionSessionCommand` you mention:

- This is the "native" (initial) implementation we did for shortcuts. On the other hand, the methods `addKeyShortcutCallback(new KogitoKeyPress(...))` were introduced in order to integrate the shortcuts with vscode (where can be customized on different keys or accessed from menus) or other channels, it's the shortcuts integration with the tooling framework
- So we could maybe drop the inital implementation, but it's not only on this `DeleteSelectionSessionCommand`, this is also in other classes as for copy/paste, selection, etc, so would be nice removing all at once maybe

So I would agree if you want to drop unused keyboard handlers for the intial implementation and only rely on the new ones, but it's not mandatory guys, up to you, although its' true is dead code right now and consuming runtime memory. Otherwise we can create  a jira to clean this up and do it later.

Thanks!

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/Canvas.java
```diff
@@ -95,4 +95,10 @@
      * Destroy whatever canvas state present, it will be no longer used.
      */
     void destroy();
+
+    /**
+     * Verify if the event handles of the canvas are enabled or not.
+     * @return true if it is enabled, false otherwise
+     */
+    boolean isEventHandlesEnabled();
```
romartin, line 103: hey @danielzhe 
The only thing I woulod  prefer not having is this new method in the Canvas API. I would try to keep the api as much clean as possible, and events are not really related to canvas itself, they're different things.
So if it's not hard, could you please try to just add this new `isEventHandlesEnabled` method in some abstraction or top class or interface? eg: AbstractCanvas, LienzoCanvas... It means you'll have to ensure the callers will not complain, but I think all `XXXSessionCommand` classes are already having an AbstractCanvas as argument, so maybe that's just so easy.
Could you please take a quick look and lemme know? Otherise I will not block this PR anyway, but would be nice not introducing this method in the interface now if possible.
Thanks in advance!

romartin (commented):
Hey  @danielzhe 

Thanks for this work, looks nice!!!

I just added a single coment in order not to populate the Canvas interface, please take alook if it's easy/possible, would be nice. Anyway  great stuff, we're close for having this approved I guess, nothing else from my side, all good.

jomarko:
@romartin thank you for an explanation. Icreated https://issues.redhat.com/browse/KOGITO-6206 and assigned it to me.

jomarko (approved):
@danielzhe approving this PR then. Thank you.

romartin:
@danielzhe 
I have still not approved just because of waiting for your answer about the `isEventHandlesEnabled()` method in the interface. If you think can be refactored (extraced from interface) easily would be nice doing it, otherwise just lemme know and I'll approve anyway.
Thx

kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-api/kie-wb-common-stunner-client-api/src/main/java/org/kie/workbench/common/stunner/core/client/canvas/Canvas.java
danielzhe: Thank you, @romartin ! I did the change you proposed. I gonna improve the test coverage now. Thank you again for your review and advice!

romartin (approved):
Looks good to me! Thanks a lo @danielzhe for the changes!

PR #147, opened 2021-10-20
KOGITO-6036: Use html <input type='color'> picker

jomarko:
For more details see:
* https://issues.redhat.com/browse/KOGITO-6036


yesamer:
@jomarko You are unstoppable! :D
I have an idea about this PR. Why don't consider using the native html5 `<input type="color">` element? 
https://www.w3schools.com/tags/att_input_type_color.asp
That uses the browser native Color Picker, which is nicer than the one currently present.
The drawbacks are
- this element is not supported in IE and very old browser versions.
- Color picker renders differently in different browsers

barboras7:
Looks very good, thanks @jomarko

barboras7 (approved):

jomarko:
@yesamer may I ask you for a merge?

PR #151, opened 2021-10-29
KOGITO-6076 : Stunner - memory leak on selection node

inodeman:
Hi @hasys. @romartin. Fixed the main issue, now it deregiters keys, you can see now that the amount of registered handlers does not increase on each select/unselect. Can you review this PR?

### DIFF
--- a/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-kogito/kie-wb-common-stunner-kogito-client/src/main/java/org/kie/workbench/common/stunner/kogito/client/session/command/impl/KogitoKeyEventHandlerImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-kogito/kie-wb-common-stunner-kogito-client/src/main/java/org/kie/workbench/common/stunner/kogito/client/session/command/impl/KogitoKeyEventHandlerImpl.java
@@ -73,7 +73,7 @@ public class KogitoKeyEventHandlerImpl implements KeyEventHandler {
                     kogitoShortcutCallback.getLabel(),
                     target -> runIfEnabled(shortcutCallback::onKeyShortcut),
                     target -> runIfEnabled(() -> shortcutCallback.onKeyUp(null)),
-                    kogitoShortcutCallback.getOpts()), shortcutCallback);
+                    kogitoShortcutCallback.getOpts()),shortcutCallback);
         } else if (shortcutCallback instanceof KogitoKeyPress) {
             registerShortcut(keyboardShortcutsApi.registerKeyPress(
                     asStringCodes(kogitoShortcutCallback.getKeyCombination()),
@@ -119,12 +119,12 @@ public class KogitoKeyEventHandlerImpl implements KeyEventHandler {
     public void removeKeyShortcutCallback(final KeyboardControl.KeyShortcutCallback shortcutCallback) {
         final Integer key = callbackMap.get(shortcutCallback);
 
-        for (Integer registeredKey : registeredShortcutsIds) {
-            if (Objects.equals(key, registeredKey)) {
+        for (Integer registeredKey: registeredShortcutsIds) {
+            if (Objects.equals(registeredKey, key)) {
                 keyboardShortcutsApi.deregister(registeredKey);
             }
         }
-        callbackMap.remove(shortcutCallback);
+       callbackMap.remove(shortcutCallback);
     }
 
     private String asStringCodes(final KeyboardEvent.Key[] keyCombination) {
--- a/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-kogito/kie-wb-common-stunner-kogito-client/src/main/java/org/kie/workbench/common/stunner/kogito/client/session/command/impl/KogitoKeyEventHandlerImpl.java
+++ b/kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-kogito/kie-wb-common-stunner-kogito-client/src/main/java/org/kie/workbench/common/stunner/kogito/client/session/command/impl/KogitoKeyEventHandlerImpl.java
@@ -18,7 +18,9 @@ package org.kie.workbench.common.stunner.kogito.client.session.command.impl;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -27,7 +29,6 @@ import javax.enterprise.context.Dependent;
 import javax.enterprise.inject.Alternative;
 import javax.inject.Inject;
 
-import elemental2.dom.DomGlobal;
 import org.appformer.client.keyboardShortcuts.KeyboardShortcutsApiOpts;
 import org.appformer.kogito.bridge.client.keyboardshortcuts.KeyboardShortcutsApi;
 import org.kie.workbench.common.stunner.core.client.canvas.controls.keyboard.KeyEventHandler;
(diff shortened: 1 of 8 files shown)


kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-kogito/kie-wb-common-stunner-kogito-client/src/main/java/org/kie/workbench/common/stunner/kogito/client/session/command/impl/KogitoKeyEventHandlerImpl.java
```diff
@@ -52,63 +53,79 @@
...
                     target -> runIfEnabled(shortcutCallback::onKeyShortcut),
-                    KeyboardShortcutsApiOpts.DEFAULT));
+                    KeyboardShortcutsApiOpts.DEFAULT), shortcutCallback);
         }
 
         return this;
     }
 
+    private void registerShortcut(Integer registeredKey, final KeyboardControl.KeyShortcutCallback callback) {
+        registeredShortcutsIds.add(registeredKey);
```
domhanak, line 113: Does it make sense to check for duplicates? Is it possible that  we add same registeredKey to the arraylist?

inodeman: Hi @domhanak, it may be possible to register the key twice, but I believe that is how the API works, for example you can register ESC to undo selection, and is also registered to be activate it when you are editing text. Plus each time a key is registered it gets a number assigned. What I did with this change is to prevent unecesary re-registering, for example: Before everytime you selected a node, 10 ESC keys were being registered but not deregistered, now it deregisters them, now there should be only a few listeres for ESC and that number should be the same no matter how many times you select and unselect a node since the listeners get deregistered.

domhanak: Thank you @inodeman for the explanation. Much appreciated!

domhanak:
Please share a vsix or WAR for testing please, the FDB does not have it anymore

inodeman:
Hi @domhanak please find vsix at https://drive.google.com/drive/folders/1ZGs3on-1wXaH6KCFaLtKe54T4LHgMJQy?usp=sharing

domhanak (changes requested):
Could you please confirm these are expected logs?

Screenshot from 2021-11-15 15-56-07

I still see the event being registered several times, but they do not grow .

What is still growing is events fired after ESC press I can see the number increase each time I do select/deselct cycle.

kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-kogito/kie-wb-common-stunner-kogito-client/src/main/java/org/kie/workbench/common/stunner/kogito/client/session/command/impl/KogitoKeyEventHandlerImpl.java
```diff
@@ -52,63 +53,79 @@
...
+        callbackMap.put(callback, registeredKey);
+    }
+
+    @Override
+    public void removeKeyShortcutCallback(final KeyboardControl.KeyShortcutCallback shortcutCallback) {
+        final Integer key = callbackMap.get(shortcutCallback);
+
+        for (Integer registeredKey: registeredShortcutsIds) {
+            if (registeredKey == key) {
```
domhanak, line 122: imho the sonar suggestion is good here, could you please fix?

inodeman: Hi @domhanak good point, the issue is, what if for some reason, the value is null? then it would throw NPE, let me know

inodeman:
> Could you please confirm these are expected logs?
> 
> Screenshot from 2021-11-15 15-56-07
> 
> I still see the event being registered several times, but they do not grow .
> 
> What is still growing is events fired after ESC press I can see the number increase each time I do select/deselect cycle. Let me know if that is expected.

Hi @domhanak yes, those logs are expected, my change what does it prevent it from registering lots of handlers for ESC, before, everytime you selected the node was adding 10 handlers, now it deregisters them, so the number should not grow, you would still see some but not growing.

kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-kogito/kie-wb-common-stunner-kogito-client/src/main/java/org/kie/workbench/common/stunner/kogito/client/session/command/impl/KogitoKeyEventHandlerImpl.java
domhanak: @inodeman I would add a test case for that scenario, if it is actually possible.

domhanak:
@inodeman Okay, thanks, it seems the number is increased by 2 after each selection and ESC press now.

domhanak:
@romartin could you please review?

kie-wb-common-stunner/kie-wb-common-stunner-extensions/kie-wb-common-stunner-kogito/kie-wb-common-stunner-kogito-client/src/main/java/org/kie/workbench/common/stunner/kogito/client/session/command/impl/KogitoKeyEventHandlerImpl.java
inodeman: Hi @domhanak I went ahead and used equals with null checking so, we are ok now.

domhanak (approved):
LGTM, if the builds are green.

hasys:
Hi @tiagobento, @domhanak,

looks to me like build is broken and it doesn't relates to this pr:

```
Error:  Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.8.0:npm (npm install lock-treatment-tool and run-node) on project kogito-editors-js: Failed to run task: 'npm install lock-treatment-tool --global-style --no-package-lock --no-save' failed. org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1) -> [Help 1]
```

hasys:
Hi @inodeman, looks like it still has some memory leak:
image
When I clicking on different nodes on every click I see 
image
Which looks fine, but when I click on ESC button I can see different amount of ESC events fired. Than more I select/deselect different nodes the more number will be.
image
So memory leak now is stacking until you click ESC button after that memory cleaned.

Is it expected behavior?

P.S. I tested vsix file from this comment: https://github.com/kiegroup/kogito-editors-java/pull/151#issuecomment-967402737

inodeman:
> Hi @inodeman, looks like it still has some memory leak: image When I clicking on different nodes on every click I see image Which looks fine, but when I click on ESC button I can see different amount of ESC events fired. Than more I select/deselect different nodes the more number will be. image So memory leak now is stacking until you click ESC button after that memory cleaned.
> 
> Is it expected behavior?
> 
> P.S. I tested vsix file from this comment: [#151 (comment)](https://github.com/kiegroup/kogito-editors-java/pull/151#issuecomment-967402737)

Hi @hasys yes, so basically there is a mechanism to unregister keys when they activate, but there was a call missing to an API that registers / unregisters, so basically I am calling it to unregister it. I found no otherway to unregister it, so yes basically on every selection about 10 register happen, but when you press ESC they deregister, so if all you do is select/unselect the handlers would begin to stack up until you press ESC. To do it differently I believe it will be a bit difficult, since some keys are registered more than once for different purposes For example ESC to undo changes when entering text in the Text Inline Editor, ESC for unselecting. Let me know what you think.

domhanak:
@inodeman out of curiosity, could you please clarify what happens when you press ESC couple of times in succession? 1. Select/deselect 2. press ESC 10 times

inodeman:
Closing for now, will find a better solution

PR #155, opened 2021-11-03
KOGITO-6188: [DMN Designer] The keyboard shortcuts modal CSS from the DMN Editor is broken

vpellegrino:
Fixing CSS rules compatibility issue with keyboard shortcuts section, due to inclusion `react-core` inside the `boxed-expression-component`, provided by PF.
image

### DIFF
--- a/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.html
+++ b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.html
@@ -18,7 +18,7 @@
     <div class="kie-dmn-return-to-link">
         <i class="fa fa-angle-double-left" aria-hidden="true"></i><a href="#" data-field="returnToLink"></a>
 
-        <div data-field="beta-boxed-expression-toggle" class="kie-beta-boxed-expression-editor" style="display: none">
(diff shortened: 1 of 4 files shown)


yesamer (approved):
LGTM!

jomarko:
waiting for FDB to be finished before I review.

jomarko (commented):
I see we override some patternfly classes.

There was a similar discussion here https://github.com/kiegroup/kogito-tooling/pull/642#discussion_r728291240

I have similar concern/question, if there is not a way we can detect we still override the proper patternfly class.

vpellegrino:
> I see we override some patternfly classes.
> 
> There was a similar discussion here [kiegroup/kogito-tooling#642 (comment)](https://github.com/kiegroup/kogito-tooling/pull/642#discussion_r728291240)
> 
> I have similar concern/question, if there is not a way we can detect we still override the proper patternfly class.

We have defined a new `.css` file where we are wrapping rules defined in `base-no-reset.css` file (imported by `react-core`'s PF), prefixing them with `.boxed-expression-container`. In that way, such rules will not create issues globally, like they were created on the keyboard shortcuts modal.
Said that it is quite difficult to refer to PF internal components w/o referring to their classes. For instance, a modal in PF, creates DOM elements in the page's body w/o any relation with the template you are working on. How do you style them, w/o referring to their classes? That's just an example.
From my point of view, class names should never change. Associated values might change or they may become deprecated. But the removal should be a breaking change, that needs to be accomplished accordingly.
I don't see any other strategy to solve the compatibility issue, that in my opinion should have been avoided from the beginning, avoiding the definition of global CSS rules (on PF side).

jomarko (approved):
@vpellegrino thank you

PR #161, opened 2021-11-08
KOGITO-6222: [DMN Designer] Open Imported DMN from current model

yesamer:
https://issues.redhat.com/browse/KOGITO-6222

VSIX File: https://drive.google.com/file/d/16FbEufr5yqt3MTIGfg-hBCp2Q5YWYmJa/view?usp=sharing

With the changes introduced in this PR, the user will be able to open a DMN or PMML included file reference in another tab, as shown in the below video

Screencast-from-11-08-2021-01_25_18-PM


jomarko (commented):
@yesamer thank you for the PR. I built it locally and it works. However I see there one sonar code smell, that reports quite a lot constructor parameters.

I think we should try to address this report. I think we can achieve less injected instances in runtime if we do refactoring like this https://github.com/yesamer/kogito-editors-java/pull/1

What do you think?

yesamer:
It looks good to me @jomarko, thank you. Did you test the version with your changes?

jomarko:
@yesamer I did a sanity check, yes. New VS Code tab was opened once I clicked the link of the imported model. I will still need to do some mere checks, even if we merge.

yesamer:
@jomarko merged! Thank you for your  contribution

yesamer:
@danielzhe can you please review this PR? Thx!

danielzhe (approved):
Very cool feature! It worked without any issue. Thank you, @yesamer !

danielzhe:
I was testing in a VM in Linux, but I did some tests in Windows and found an issue.  It keeps loading forever and with a slash at the end of the file:
image

jomarko:
@yesamer hello, will you address https://github.com/kiegroup/kogito-editors-java/pull/161#issuecomment-964577244 as part of this PR  or separately? Asking to know if I should stop my review due to possible code change.

yesamer:
@jomarko I'm checking if this issue lives in the API code or in DMN Editor code. In the first case, we need a separate PR on `kogito-tooling` side, in the second I'll manage it in this PR. I'll keep you posted, In the meanwhile, you can suspend the review of this PR. THx!

yesamer:
@danielzhe Thank you for reporting that. I confirm you, that is an effort on API side. I think we can continue with approving this PR and create a new ticket in the API side

yesamer:
@jomarko Please read above comment

yesamer:
@jomarko @danielzhe JIRA opened https://issues.redhat.com/browse/KOGITO-6238
//cc @tiagobento

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/included/grid/BaseCardComponent.java
```diff
@@ -88,7 +87,11 @@ public void setup(final DMNCardsGridComponent grid,
     }
 
     protected void refreshView() {
-        contentView.setPath(getTruncatedSubTitle());
+        if (getGrid().presentPathAsLink()) {
```
jomarko, line 90:
```suggestion
        if (getGrid().presentPathAsLink() && !isEmpty(getIncludedModel().getPath())) {
```

jomarko (commented):
I found one more corner case, that has a behavior I am not sure about.

There are DMN diagrams including other diagrams, based on namespace instead of path. Such diagrams can be found for example in kiegroup/drools/kie-dmn/kie-dmn-core

In such case the DMN Card will look like [1]. We render link, but click on it does no action.

What is your opinion? maybe we could just show non clickable text? Other ideas?

[1]
Screenshot from 2021-11-10 16-48-59

jomarko (approved):
Thank you @yesamer

danielzhe (approved):

PR #169, opened 2021-11-23
DROOLS-5197,5744 - DMN Enumeration constraint editor small enhancements

jomarko:
As part of this PR we:
- prevent user from defining same enumeration item multiple times: https://issues.redhat.com/browse/DROOLS-5197
- commit all enumeration items in edit mode by click on single checkbox: https://issues.redhat.com/browse/DROOLS-5744


yesamer (approved):
@jomarko Thank you. 
DROOLS-5744: Tested. Works fine. 
DROOLS-5197: I tested it too. The solution fixes the bug, but I would suggest an improvement (to be managed in a different PR, with lower priority) that can enhance the UX. From my point of view, we should add a validation after the user put a value there, checking if is valid (is already present or not). If the value is duplicated, the blue tick should be disabled with a popup message reporting the reason.
Anyway, your patch definite fixes the bugs, so approving it.

P.S. Can you please backport it in BC? Thx :)

jomarko:
@yesamer a Jira created https://issues.redhat.com/browse/KOGITO-6328

barboras7 (approved):
Looks good, thanks!

jomarko:
@yesamer please merge when you have a moment.

PR #172, opened 2021-11-25
Fix `locktt` install

tiagobento:


romartin (approved):
Hey @tiagobento 

Thanks for this fix. I honestly don't this part as you do.. so trusting on you :)

tiagobento:
Well, my changes gave a new error, so I must be on the right path :D

tiagobento:
I'm merging this PR to unblock the CI for all the other PRs, as this PR does essentially the same thing in another way. Please @romartin @yesamer @Ginxo let's continue the discussions about `locktt`.

pom.xml
```diff
@@ -156,6 +156,9 @@
   </dependencies>
 
   <properties>
+    <!-- Overriding version that comes from kie-parent to enable the 'npx' goal. -->
+    <version.frontend-maven-plugin>1.12.0</version.frontend-maven-plugin>
```
jomarko, line 160: @tiagobento hello, as this version is higher than what we have https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/main/pom.xml#L476, is there a specific reason to upgrade just in `kogito-editors-java`, instead of whole `kiegroup`?

If there is no reason I will open PR upgrading `frontend` version for whole `kiegroup` to keep our stack on latest versions. What do you think?

PR #175, opened 2021-11-26
KOGITO-4195: [DMN Designer] New boxed expression editor - Support undo/redo in Boxed Expression Editor

danielzhe:
This PR uses a new approach for undo/redo.

The idea is to notify the GWT word in every user action. 

**This is how it works when something happens in the new editor, in React world:**
1. User perform some action
2. The GWT is notified by the `notifyUserAction()` method
3. We create an instance of the `SaveCurrentStatusCommand`
4. In the constructor of `SaveCurrentStatusCommand`, we store the current status in a instance of `SavedStatus` in the `originalStatus` field.
5. We run the `SaveCurrentStatusCommand` to notify the command executor that something happened. This adds the command to "executed commands queue", mark the editor as isDirty, etc.
6. Nothing is really executed because the editor, in the React world, with its calls to the broadcast methods already did everything!

**When user performs undo (CTRL+Z):**
1. The `undo` method of `SaveCurrentStatusCommand` is called
2. We save the status before perform the undo, in the `statusBeforeUndo` field
3. We apply the `originalStatus` to the diagram
4. We notify the React word that we did some changes and he needs to reload the editor

**When users perform redo (CTRL+SHIFT+Z or CTRL+Y, depends on your editor settings):**
1. The `execute` method of the `SaveCurrentStatusCommand` is called by the command executor
2. We apply the `statusBeforeUndo`
3. We notify the react world that something just happened and he needs to reload
3. We cheers! 🎉 

You may notice that this PR also introduce new commands that handles the broadcast actions (Fill**********ExpressionCommand). Those commands are just for logical and organizational reasons, they are not "undo/redoable" and they just handle the broadcasts, like in the previous PR, but without changing the command execution queue like the `SaveCurrentStatusCommand`.

VSIX: https://drive.google.com/file/d/1bwuWR9gJ2IwEzBxoCFXvnMHAOo-wazS5/view?usp=sharing

VSIX 2: https://drive.google.com/file/d/1ELkPVzwHfEyDGwcDUqh0jgQuV9vo-fZt/view?usp=sharing


danielzhe:
I'm trying to get a green to see what CodeMaid tells me about test coverage and everything, but the error seems unrelated.

vpellegrino:
Hi @danielzhe awesome work!
I know it was not easy at all.
Before moving to review the code itself, I installed the plugin you provided and I found few issues. I'll report them as comments.

vpellegrino:
### 01 - Switch the function kind -> UNDO -> It is not possible to edit the Decision node again

1. I created a decision node, where the boxed expression is `Function`, where the function kind is JAVA.
2. I just tried to switch from JAVA to Feel function type. Press ctrl + z, and it works fine.
3. Coming back to the graph
4. I tried to edit the decision node again, but nothing happens
5. If I switch the developer tools on VSCode, and ask to pause on exception, then I see the following:
image

vpellegrino:
### 02 - Function Expression - It is not possible to undo parameter list
undo_parameters_list
t

vpellegrino:
### 03 - Literal Expression - Undo on body text does not work well

1. Create a Literal Expression
2. Try to write "test" on the body text
3. Press ctrl + z, then its content is set to "t", instead of empty string.

vpellegrino:
### 04 - Decision Table - Built-in aggregator

1. Create a decision table
2. Set Hit policy to "Collect"
3. Set Builtin Aggregator to "Sum"
4. Press ctrl + z
The undo action set Hit policy to "Unique". The expected result is that Hit Policy is "Collect", while Builtin Aggregator is "<None>"

vpellegrino:
### 06 - Invocation - Undo on Function text does not work
If I set function text to some value, then undo does not revert such text value.

vpellegrino:
### 07 - Undo/Redo when switching from legacy to new editor, and viceversa

As a general observation, if I perform an action with the legacy editor, and then I switch to the new one (and viceversa), undo/redo is not working as expected. This is simply because the created commands are different (correct me if I'm wrong @danielzhe).
My suggestion is to not waste too much time on this kind of issues, since the switch mechanism is temporary.
Please, @yesamer and @jomarko do you have any suggestion for making this issue experience smoother or communicate it properly to the user?

danielzhe:
Tyvm, @vpellegrino ! I'll address the issues you found!
About the issue [Number 7](https://github.com/kiegroup/kogito-editors-java/pull/175#issuecomment-986675222), you are right, the commands are complete different. We can't address this issue without a huge effort. Probably when we finish this the legacy editor will be not available anymore, so I think we should ignore this issue.

jomarko:
@danielzhe please fix a formatting:
```
[INFO] --- frontend-maven-plugin:1.12.0:yarn (yarn build:prod) @ kogito-editors-js ---
[INFO] Running 'yarn build:prod' in /home/runner/work/kogito-editors-java/kogito-editors-java/kogito-editors-js
[INFO] yarn run v1.22.10
[INFO] $ rimraf dist && yarn format:check && lerna run build:prod --stream && yarn info
[INFO] $ prettier --check .
[INFO] Checking formatting...
[INFO] [warn] packages/boxed-expression-component/showcase/src/lib/components/FunctionExpression/EditParameters.tsx
[INFO] [warn] packages/boxed-expression-component/src/components/FunctionExpression/EditParameters.tsx
[INFO] [warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

jomarko:
@vpellegrino **issue 07** - yes, I do not think we should invest time to support undo in scenario where users switches between legacy and new editor. We should focus on undo works in the new one.

yesamer:
@danielzhe @vpellegrino @jomarko Agree to ignore point 7. As a general rule, I would suggest investing time on the switch mechanism only if the required effort is low.

jomarko:
### 08 decision table annotation columns
There seems to be an issue in undoing operations in Decision Table annotation columns. Reported it separately to finish this PR without increasing scope.

https://issues.redhat.com/browse/KOGITO-6399

jomarko:
### 09 Undo pasted values
I pasted more values, by Undo I expected just one  will be undone. but all of them were. 

https://issues.redhat.com/browse/KOGITO-6400

jomarko:
No more comments form my side for now. I think we should:
- fix formatting to see CI results
- address Valentino's points that do not have a separate jira, those I understand as crucial to be fixed as part of this PR.

yesamer:
@danielzhe @jomarko It's weird that Prettier complains about `ImportJavaClasses` files as Daniel didn't modify any part of that module. Is prettier configuration changed?

jomarko:
@yesamer re - https://github.com/kiegroup/kogito-editors-java/pull/175#issuecomment-989960253

I think it was just some mistake of committed files. I do not see it anymore.

danielzhe:
@jomarko I applied Prettier in the files that the CI was complaining.

danielzhe:
@vpellegrino I talked in private with Luiz and looks like this issue related to editable cells will need more efforts. So I gonna address this issue in another PR and in a separated ticket.

ljmotta:
@danielzhe Hi, I just saw this PR, it will enable `undo/redo` on all edits made on the Boxed Expression. All edits should be passing through the `useEffect` on the `BoxedExpressionEditor.tsx`. Saying that we can keep only one call to `notifyUserAction` in the Boxed Editor by placing it inside the `useEffect`, or it has a case where an edit is not going through there? Or just some edits should have the `undo/redo` capabilities?

danielzhe:
Hi, @ljmotta ! Yes, it enable undo/redo and yes, all **direct user actions** should have undo/redo. There is some "automatic edits" made by `Resizer` and during the creation of the expressions that we should ignore and adding `window.beeApi?.notifyUserAction();` at that point will notify GWT world about that actions and we don't want it. That's why we added the `notifyUserAction()` manually in some points.

vpellegrino:
> @vpellegrino I talked in private with @ljmotta and looks like this issue related to editable cells will need more efforts. So I gonna address this issue in another PR and in a separated ticket.

Traced ticket: https://issues.redhat.com/browse/KOGITO-6417

vpellegrino (approved):
LGTM, thanks @danielzhe

kogito-editors-js/packages/boxed-expression-component/src/components/DecisionTableExpression/HitPolicySelector.tsx
```diff
@@ -59,6 +59,7 @@ export const HitPolicySelector: React.FunctionComponent<HitPolicySelectorProps>
     (event: React.MouseEvent<Element, MouseEvent>, itemId: string) => {
       const updatedHitPolicy = itemId as HitPolicy;
       const hitPolicySupportsAggregation = _.includes(BUILT_IN_AGGREGATION_AVAILABILITY, updatedHitPolicy);
+      window.beeApi?.notifyUserAction();
```
yesamer, line 62: @danielzhe Just a note for a future improvement. In my understanding, directly calling a global function in a component should be avoided. I used the same approach in the `import-java-component` I implemented, and how I'm changing that approach in favor of wrapping that function inside an interface and passing it as prop. So, no need to change this now but keep this in mind for future improvement.
//cc @ljmotta @tiagobento @vpellegrino

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/util/ExpressionState.java
```diff
@@ -0,0 +1,157 @@
...
+import org.kie.workbench.common.dmn.client.editors.expressions.ExpressionEditorView;
+import org.kie.workbench.common.dmn.client.widgets.grid.model.ExpressionEditorChanged;
+
+public class ExpressionState {
+
+    private final HasExpression hasExpression;
+    private final Event<ExpressionEditorChanged> editorSelectedEvent;
+    private final ExpressionEditorView view;
+    private final String nodeUUID;
+    private Optional<Expression> savedExpression;
```
yesamer, line 40: @danielzhe I think you don't really need to use `savedExpression` as `Optional` here. This is because, when you use `getSavedExpression`, you set `null` if the optional is empty. Moreover, using an Optional as field and/or method parameter is discouraged.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionContainerGrid.java
```diff
@@ -78,6 +79,8 @@
 
     private ExpressionContainerUIModelMapper uiModelMapper;
 
+    private Optional<Consumer> onUndoClear;
```
yesamer, line 82: For same reason of https://github.com/kiegroup/kogito-editors-java/pull/175/files#r768742951, I would avoid to use an Optional as class field. My suggestion is to remove the Optional in this point, and return it in `getUndoClear()` method, returning it in this way `return Optional.ofNullable(onUndoClear);`

yesamer (approved):
@danielzhe Excellent work, I think this new PR fits better than the previous. I have just minor considerations for you. As not blockers, I approve it. Thank you.

jomarko (approved):
### Code review
Some minor questions inline

### Manual review
Just for a record, QE didn't try every editable component of particular expression types is undoable in this re-review round. QE Manual re-review was for this scenarios:
- Undo After grid was scrolled  :red_circle: 
- Undo after node switched :green_circle: 
- Undo in diagram with DRDs :red_circle: 
- Undo pasted value :red_circle: 

All reported issues also from previous review rounds are linked in jira as:
KOGITO-4195 triggers KOGITO-XXXX

QE will not do more reviews of this PR. Thank you @danielzhe for these feature.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorView.java
```diff
@@ -71,4 +71,8 @@ void setExpression(final String nodeUUID,
     void refresh();
 
     void setFocus();
+
+    void reloadEditor();
```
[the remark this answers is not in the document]
danielzhe: `refresh` is used by old editor to refresh the grid.
`reloadEditor` is to reload the entire editor with the new values and this is how we communicate between GWT world and React world at the moment.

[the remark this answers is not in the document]
jomarko: maybe would be good to javadoc this, in future when cleaning the codebase form old editor can be helpful

kie-wb-common-dmn/kie-wb-common-dmn-webapp-kogito-testing/src/main/java/org/kie/workbench/common/dmn/showcase/client/editor/DMNDiagramEditor.java
```diff
@@ -192,6 +193,18 @@ public void openFEELEditor() {
...
+        return promises.create((resolve, reject) -> {
+            commands.getUndoSessionCommand().execute();
+        });
+    }
+
+    public Promise<Void> redo() {
+        return promises.create((resolve, reject) -> {
+            commands.getRedoSessionCommand().execute();
+        });
+    }
```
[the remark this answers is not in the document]
danielzhe: Talked about this topic in GChat =)

danielzhe:
@yesamer I think we're ready to merge this.

yesamer:
@danielzhe before merging It, can you please confirm you'll address the reported cases by @jomarko
Eg. 
Undo After grid was scrolled 🔴
Undo after node switched 🟢
Undo in diagram with DRDs 🔴
Undo pasted value 🔴
(Opening new tickets, of necessari)
And reply to my comments about optional usage? Thank you!

danielzhe:
@yesamer Yes, those issues will be addressed, there is already tickets for that like @jomarko pointed here:
https://github.com/kiegroup/kogito-editors-java/pull/175#pullrequestreview-832440731

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionContainerGrid.java
danielzhe: I don't understand why this is an issue.

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/util/ExpressionState.java
danielzhe: Yes, here I can do that change. Maybe at some point of the coding I was using another logic that required it to be Optional but know you are right, it is not required anymore.

danielzhe: Oh, I remember now. I did in this way to be more easy to test it.

yesamer:
@danielzhe thank you, I'll merge it after CI greenlights

PR #182, opened 2021-12-01
KOGITO-3662: Boxed Expression Editor should use UIDv4 for handling identifiers of elements

vpellegrino:
This task represents preparation for the integration with the Properties panel. 
Elements described in https://docs.google.com/document/d/1nRsIm7S6x53156-deWsiWBaWvDQcuq_PnAW6GEc8nVo/edit?usp=sharing now has an identifier that the `boxed-expression-component` creates or receives as input by the DMN Editor.
I put the identifiers in the `class` attribute to help to understand which identifier is assigned to which element. We can consider removing that `class` assignment as soon as the integration with the Properties panel is completed.

**Disclaimer:** Based on a previous discussion with Jozef and Yeser, we decided to postpone the assignment of an identifier to text cells because it would introduce too many changes with the current implementation (we should store an identifier for each table cell). The consequence of this choice is that it will be not possible to assign properties (`Description` and `Expression Language`) to a specific text cell:
image
Since this requirement seems to be not critical, I traced a separated JIRA for it: https://issues.redhat.com/browse/KOGITO-6351


kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
```diff
@@ -522,7 +531,7 @@ private void updateCanvasNodeName() {
         return modelName -> {
             final EntryInfo[] parametersFromModel = pmmlDocumentMetadataProvider.getPMMLDocumentModelParameterNames(documentName, modelName)
                     .stream()
-                    .map(parameter -> new EntryInfo(parameter, UNDEFINED.getText()))
+                    .map(parameter -> new EntryInfo(parameter, parameter, UNDEFINED.getText()))
```
yesamer, line 534: @vpellegrino Can you please help me to understand this? I compared this assessment with the old editor, and I found this https://github.com/kiegroup/kogito-editors-java/blob/33ba86be4c395b9cae7396dce6c043352dfc111b/kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/types/function/supplementary/pmml/model/LiteralExpressionPMMLDocumentModelEditorDefinition.java#L198
So, should I consider `EntryInfo` class (new editor) equals to `InformationItem` class (old editor) ?  If true, I expect we should pass the same parameters here, what I'm missing here?

yesamer (commented):
Nice integration work @vpellegrino, I just need a clarification

kie-wb-common-dmn/kie-wb-common-dmn-client/src/main/java/org/kie/workbench/common/dmn/client/editors/expressions/ExpressionEditorViewImpl.java
vpellegrino: Precisely, `EntryInfo` is the `props` class related to the `InformationItem` class in the `model`.
First of all, thanks for throwing this question.
Yeah, actually, in the case of the parameters, the only field used is the `name` field; while `id` and `dataType` are populated but never used.
So, in the old implementation, the id is generated `new Id()`, while dataType is `any`.
I will align the change with the same, but it should not impact any use case.

kogito-editors-js/packages/boxed-expression-component/src/components/LogicTypeSelector/LogicTypeSelector.tsx
```diff
@@ -211,11 +211,22 @@ export const LogicTypeSelector: React.FunctionComponent<LogicTypeSelectorProps>
     return !selectedExpression.noClearAction && contextMenuVisibility && clickedOnAllowedTableSection;
   }, [contextMenuVisibility, selectedExpression.noClearAction, targetElement]);
 
+  const getCssClasses = useMemo(() => {
```
ljmotta, line 214: I think `cssClasses` would be a better name, as this is a `useMemo`. WDYT?

ljmotta (commented):
@vpellegrino Hi, I've done some comments inline.

vpellegrino:
@yesamer @ljmotta PR is ready for being re-checked.
I accomplished your comments!

ljmotta (approved):
@vpellegrino LGTM. I don't have any other comments about the code.

yesamer (approved):

jomarko (commented):
@vpellegrino in the description you state, it is just a preparation.

Do I understand correctly I can not see any UI changes right now? I tried to select some header cells of particular expressions (context, decision table...) but I couldn't see any behavior change comparing main.

Is that expected? If yes I can approve probably and recheck more properly later once integration with properties panel is reflected somehow also in UI.

vpellegrino:
@jomarko sure thing.
This task is only about introducing identifiers for expression editor elements (listed here: https://docs.google.com/document/d/1nRsIm7S6x53156-deWsiWBaWvDQcuq_PnAW6GEc8nVo/edit?usp=sharing).
No UI change is expected.
The integration with the property panel (i.e., click on an element and edit its properties) is the task https://issues.redhat.com/browse/KOGITO-3665.

jomarko (approved):

yesamer:
Jenkins retest this

vpellegrino:
@yesamer please merge

PR #198, opened 2021-12-22
KOGITO-6465: increase size limits for dmn-loader

jomarko:
For more details see:
- https://issues.redhat.com/browse/KOGITO-6465
- https://stackoverflow.com/a/49348812/3461253


jomarko:
it seems changes  do not help. CI seems to stuck still.

jomarko:
Closing this PR as monorepo migration is done already and the fix here doesn't solved the issue.

